PR #11, opened 2021-11-17
Add CORS support

wraitii:
- Adds a dependency on [Flask-CORS](https://flask-cors.readthedocs.io/en/latest/) to be able to answer XHR calls on localhost.
- Update poetry.lock accordingly

I've tested this locally FWIW

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,6 +1,6 @@
 [[package]]
 name = "aiohttp"
-version = "3.8.0"
+version = "3.8.1"
 description = "Async http client/server framework (asyncio)"
 category = "main"
 optional = false
@@ -47,7 +47,7 @@ tests = ["pytest", "pytest-asyncio", "mypy (>=0.800)"]
 
 [[package]]
 name = "async-timeout"
-version = "4.0.0"
+version = "4.0.1"
 description = "Timeout context manager for asyncio programs"
 category = "main"
 optional = false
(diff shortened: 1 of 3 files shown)


FabijanC:
Could you describe here how you tested it?

wraitii:
Essentially by running the Briq project in browser. The problem is that you need a browser to enforce CORS, so I don't see an easy way to test it programmatically.

The basic idea is that I tested `fetch("http://localhost:5000/feeder_gateway/get_contract_addresses")` to return a `501 NOT IMPLEMENTED` instead of a CORS-related error.

FabijanC:
First of all, I recommend targeting `http://localhost:5000/is_alive`.

I'm aware of needing a browser to test CORS, so I usually test it by opening a website and running a `fetch` in the browser console. Anyway, as expected, most sites don't allow such behavior due to their `Content-Security-Policy`, so my question is - where would you be using this? In custom websites that would such requests?

Also, I'm not sure what you mean when you say "running the Briq project".

wraitii:
> I'm aware of needing a browser to test CORS, so I usually test it by opening a website and running a `fetch` in the browser console. Anyway, as expected, most sites don't allow such behavior due to their `Content-Security-Policy`, so my question is - where would you be using this? In custom websites that would allow such requests?

The issue is when a website wants to perform a `fetch` request directly to the devnet, as one would the alpha net. That request is blocked by CORS because the server (the devnet) is not answering CORS requests right now, and so the `fetch` call fails. I believe whether the source website has a content security policy or not does not come into play.

> Also, I'm not sure what you mean when you say "running the Briq project".
i'm referring to http://github.com/S-L-Technologies/brq-builder, which is my WIP project. It's basically just a SPA.

FabijanC:
> The issue is when a website wants to perform a `fetch` request directly to the devnet, as one would the alpha net. That request is blocked by CORS because the server (the devnet) is not answering CORS requests right now, and so the `fetch` call fails.

Yes, I understand that, this is what your PR addresses.

> I believe whether the source website has a content security policy or not does not come into play.

No need to believe, try it out!

FabijanC:
For the record, @wraitii and I did some chatting and agreed to merge this.

PR #21, opened 2021-12-20
Support linting; Add plugin-venv test

FabijanC:
## Usage
- This PR contains no usage changes.

## Dev
- Adds a linting step to the CI/CD workflow:
  - Performed with `pylint` and configured through `.pylintrc`.
  - Make necessary changes to conform to the standards set by the linter.
- Add testing with plugin using a venv:
  - So far only the dockerized plugin version was tested.
  - Dockerized and venv test are specified as separate steps in job `test`.
  - ~~Rely on hardhat's --config option for specifying which config file to use.~~
- Instead of running all tests in `starknet-hardhat-example`, run only `quick-test.ts`.

### DIFF
--- a/starknet_devnet/starknet_wrapper.py
+++ b/starknet_devnet/starknet_wrapper.py
@@ -5,14 +5,15 @@ starkware.starknet.testing.starknet.Starknet.
 
 import time
 from copy import deepcopy
-from typing import List, Dict
+from typing import Dict
+from starkware.starknet.business_logic.internal_transaction import InternalDeploy
 
 from starkware.starknet.business_logic.state import CarriedState
+from starkware.starknet.services.api.gateway.transaction import InvokeFunction
 from starkware.starknet.testing.starknet import Starknet
-from starkware.starknet.services.api.contract_definition import ContractDefinition
-from starkware.starknet.testing.state import CastableToAddressSalt
 from starkware.starknet.definitions.error_codes import StarknetErrorCode
 from starkware.starknet.definitions.transaction_type import TransactionType
+from starkware.starkware_utils.error_handling import StarkException
 
 from .util import Choice, StarknetDevnetException, TxStatus, fixed_length_hex
 from .contract_wrapper import ContractWrapper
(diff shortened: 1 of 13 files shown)


starknet_devnet/starknet_wrapper.py
```diff
@@ -230,15 +228,17 @@ async def store_transaction(self, contract_address: str, status: TxStatus, error
...
-            block_hash, block_number = await self.generate_block(transaction["transaction"])
+            block_hash, block_number = await self.__generate_block(transaction["transaction"])
             transaction["block_hash"] = block_hash
             transaction["block_number"] = block_number
 
-        self.transactions.append(transaction)
+        self.__transactions.append(transaction)
         return hex_new_id
 
+    # pylint: disable=too-many-arguments
```
dribeiro-ShardLabs, line 238: Maybe the argument could be a single Object/Dataclass with all these arguments as properties

starknet_devnet/starknet_wrapper.py
```diff
@@ -247,8 +247,12 @@ async def store_deploy_transaction(self, contract_address: str, calldata: List[i
             contract_address_salt=hex(salt)
         )
 
-    async def store_invoke_transaction(self, contract_address: str, calldata: List[int], entry_point_selector: int, status: TxStatus, error_message: str=None):
-        return await self.store_transaction(
+    # pylint: disable=too-many-arguments
```
dribeiro-ShardLabs, line 250: Maybe the argument could be a single Object/Dataclass with all these arguments as properties

starknet_devnet/starknet_wrapper.py
```diff
@@ -89,45 +99,30 @@ async def deploy(self, contract_definition: ContractDefinition, contract_address
             constructor_calldata=constructor_calldata,
             contract_address_salt=contract_address_salt
         )
-        await self.update_state()
+        await self.__update_state()
 
-        self.address2contract_wrapper[contract.contract_address] = ContractWrapper(contract, contract_definition)
+        self.__address2contract_wrapper[contract.contract_address] = ContractWrapper(contract, contract_definition)
 
+    # pylint: disable=too-many-arguments
```
dribeiro-ShardLabs, line 106: Maybe the argument could be a single Object/Dataclass with all these arguments as properties

dribeiro-ShardLabs (approved):

PR #24, opened 2021-12-28
Calculate transaction hashes properly

dribeiro-ShardLabs:
- Calculate transaction hashes properly
 - Rebased with previous PRs

### DIFF
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -63,7 +63,7 @@ async def add_transaction():
             error_message = err.message
             status = TxStatus.REJECTED
 
-        transaction_hash = await starknet_wrapper.store_invoke_transaction(
+        transaction_hash = await starknet_wrapper.store_wrapper_transaction(
             transaction=transaction,
             status=status,
             execution_info=execution_info,
--- a/starknet_devnet/starknet_wrapper.py
+++ b/starknet_devnet/starknet_wrapper.py
@@ -9,7 +9,8 @@ from typing import Dict
 
 from starkware.starknet.business_logic.internal_transaction import InternalDeploy
 from starkware.starknet.business_logic.state import CarriedState
-from starkware.starknet.services.api.gateway.transaction import InvokeFunction
+from starkware.starknet.definitions.transaction_type import TransactionType
+from starkware.starknet.services.api.gateway.transaction import InvokeFunction, Transaction
 from starkware.starknet.testing.starknet import Starknet
 from starkware.starknet.testing.objects import StarknetTransactionExecutionInfo
 from starkware.starkware_utils.error_handling import StarkException
@@ -17,7 +18,7 @@ from starkware.starkware_utils.error_handling import StarkException
 from .origin import Origin
 from .util import Choice, StarknetDevnetException, TxStatus, fixed_length_hex, DummyExecutionInfo
 from .contract_wrapper import ContractWrapper
-from .transaction_wrapper import TransactionWrapper, DeployTransaction, InvokeTransaction
+from .transaction_wrapper import TransactionWrapper, DeployTransactionWrapper, InvokeTransactionWrapper
 
 class StarknetWrapper:
     """
@@ -37,7 +38,7 @@ class StarknetWrapper:
         self.__hash2block = {}
         """Maps block hash to block."""
 
-        self.__num2block = {}
+        self.__num2block: Dict[int, Dict] = {}
         """Maps block number to block (one transaction per block); holds only own blocks."""
 
         self.__starknet = None
@@ -111,7 +112,7 @@ class StarknetWrapper:
             error_message = err.message
             status = TxStatus.REJECTED
 
-        transaction_hash = await self.__store_deploy_transaction(
+        transaction_hash = await self.store_wrapper_transaction(
             transaction,
             status=status,
             execution_info=DummyExecutionInfo(),
@@ -146,9 +147,10 @@ class StarknetWrapper:
     def get_transaction_status(self, transaction_hash: str):
         """Returns the status of the transaction identified by `transaction_hash`."""
 
-        if transaction_hash in self.__transaction_wrappers:
+        tx_hash_int = int(transaction_hash,16)
+        if tx_hash_int in self.__transaction_wrappers:
 
-            transaction_wrapper = self.__transaction_wrappers[transaction_hash]
+            transaction_wrapper = self.__transaction_wrappers[tx_hash_int]
 
             transaction = transaction_wrapper.transaction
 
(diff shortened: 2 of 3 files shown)


FabijanC (changes requested):
As discussed in the call:
- Make the changes required by pylint.
- Use the existing starknet config.
- Store transaction wrappers in a dict.
- Expand the TransactionWrapper class (potentially changing its constructor and properties) and have inherited classes of it (for deploy and invoke transactions).

starknet_devnet/transaction_wrapper.py
```diff
@@ -2,11 +2,102 @@
...
+
+        self.receipt = {
+            "execution_resources": execution_info.call_info.cairo_usage,
+            "l2_to_l1_messages": execution_info.l2_to_l1_messages,
+            "status": self.transaction["status"],
+            "transaction_hash": self.transaction_hash,
+            "transaction_index": 0 # always the first (and only) tx in the block
+        }
+
+    def set_transaction_failure(self, failure_key: str, error_message: str):
```
FabijanC, line 48: Don't both classes use the same failure_key, can't it be a local variable of the method instead of a function argument?

starknet_devnet/transaction_wrapper.py
```diff
@@ -2,11 +2,102 @@
...
+        """Creates a new entry `failure_key` in the transaction object with the transaction failure reason data."""
+
+        self.transaction[failure_key] = self.receipt[failure_key] = {
+                "code": StarknetErrorCode.TRANSACTION_FAILED.name,
+                "error_message": error_message,
+                "tx_id": self.transaction["transaction"]["transaction_hash"]
```
FabijanC, line 54: Can you use self.transaction_hash for this?

starknet_devnet/transaction_wrapper.py
```diff
@@ -2,11 +2,102 @@
...
+
+
+class InvokeTransaction(TransactionWrapper):
+    """Class for Invoke Transaction."""
+
+    def __init__(self, internal_transaction, status, starknet):
+
+        super().__init__()
+
+        self.transaction_hash = hex(calculate_transaction_hash(
```
FabijanC, line 89: Could you now use the internal_transaction.calculate_hash method?

starknet_devnet/transaction_wrapper.py
```diff
@@ -2,11 +2,102 @@
...
+        """Creates a new entry `failure_key` in the transaction object with the transaction failure reason data."""
+
+        self.transaction[failure_key] = self.receipt[failure_key] = {
+                "code": StarknetErrorCode.TRANSACTION_FAILED.name,
+                "error_message": error_message,
+                "tx_id": self.transaction["transaction"]["transaction_hash"]
+        }
+
+
+class DeployTransaction(TransactionWrapper):
```
FabijanC, line 58: I'd call this DeployTransactionWrapper and analogously for Invoke

starknet_devnet/starknet_wrapper.py
```diff
@@ -53,13 +31,13 @@ def __init__(self, origin):
...
-        self.__transaction_wrappers: List[TransactionWrapper] = []
-        """A chronological list of transaction_wrappers."""
+        self.__transaction_wrappers: Dict[int, TransactionWrapper] = {}
+        """Maps transaction hash to transaction wrapper."""
 
         self.__hash2block = {}
         """Maps block hash to block."""
 
-        self.__own_blocks = {}
+        self.__num2block = {}
```
FabijanC, line 40: You can put Dict[int, Dict] here (I know Dict is very wild card, but that's better than nothing)

starknet_devnet/starknet_wrapper.py
```diff
@@ -266,64 +245,49 @@ def get_block_by_number(self, block_number: int):
...
-            await self.__generate_block(transaction, receipt)
+            await self.__generate_block(transaction_wrapper.transaction, transaction_wrapper.receipt)
 
-        self.__transaction_wrappers.append(TransactionWrapper(transaction, receipt))
-        return hex_new_id
+        self.__transaction_wrappers[transaction_wrapper.transaction_hash] = transaction_wrapper
+
+        return transaction_wrapper.transaction_hash
 
     async def __store_deploy_transaction(self, transaction: InternalDeploy, status: TxStatus,
```
FabijanC, line 268: Can __store_deploy_transaction, __store_invoke_transaction and __store_transaction now be the same function? Because the two first mentioned are very similar (only differing in the wrapper class used - so it could be wrapped outside and passed to the new "merged" function)

starknet_devnet/starknet_wrapper.py
```diff
@@ -165,16 +143,15 @@ async def call_or_invoke(self, choice: Choice, specifications: InvokeFunction):
...
         return { "result": adapted_result }, execution_info
 
-    def __is_transaction_hash_legal(self, transaction_hash_int: int) -> bool:
-        return 0 <= transaction_hash_int < len(self.__transaction_wrappers)
-
     def get_transaction_status(self, transaction_hash: str):
         """Returns the status of the transaction identified by `transaction_hash`."""
 
-        transaction_hash_int = int(transaction_hash, 16)
+        if transaction_hash in self.__transaction_wrappers:
```
FabijanC, line 149: Here transaction_hash is a string I assume, but __transaction_wrappers is Dict[int, TransactionWrapper]. The initial implementation had ints as keys because I wanted to avoid issues with leading zeros (e.g. "0x01" and "0x1" not being the same hash), so that's why I did all the conversions to int.

FabijanC (changes requested):

starknet_devnet/transaction_wrapper.py
```diff
@@ -45,54 +43,43 @@ def generate_receipt(self, execution_info):
...
         super().__init__()
 
-        self.transaction_hash = hex(calculate_transaction_hash(
-            TransactionType.INVOKE_FUNCTION,
-            contract_address=internal_transaction.contract_address,
-            entry_point_selector=internal_transaction.entry_point_selector,
-            calldata=internal_transaction.calldata,
-            chain_id=starknet.state.general_config.chain_id.value,
-        ))
+        self.transaction_hash = hex(internal_transaction.calculate_hash(starknet.state.general_config))
```
FabijanC, line 82: the `__init__` method probably doesn't need the whole starknet object

starknet_devnet/starknet_wrapper.py
```diff
@@ -256,35 +260,26 @@ def get_block_by_number(self, block_number: int):
...
         else:
             await self.__generate_block(transaction_wrapper.transaction, transaction_wrapper.receipt)
 
-        self.__transaction_wrappers[transaction_wrapper.transaction_hash] = transaction_wrapper
+        self.__transaction_wrappers[int(transaction_wrapper.transaction_hash,16)] = transaction_wrapper
 
         return transaction_wrapper.transaction_hash
 
-    async def __store_deploy_transaction(self, transaction: InternalDeploy, status: TxStatus,
+    async def store_wrapper_transaction(self, transaction: Transaction, status: TxStatus,
```
FabijanC, line 271: There should be just one public method called `store_transaction`.

starknet_devnet/server.py
```diff
@@ -63,7 +63,7 @@ async def add_transaction():
             error_message = err.message
             status = TxStatus.REJECTED
 
-        transaction_hash = await starknet_wrapper.store_invoke_transaction(
+        transaction_hash = await starknet_wrapper.store_wrapper_transaction(
```
FabijanC, line 66: This should take care of its transaction storing the same way `deploy` does it. Also, `deploy` should accept the original `Transaction` object, then convert it internally to `InternalDeploy`. That way there'd be no need to access `state.general_config` or even `starknet.state` (the get_state method might not even be needed).

FabijanC (approved):
There are some propositions we should deal with in future. But looks functional, so can be merged.

PR #28, opened 2022-01-12
Fix tx storing; ignore duplicate

FabijanC:
## Usage
- Prevent 500 Internal Server Error if deployment fails:
  - Attempt to store tx only if deploy succeeds.
- Ignore duplicate deploys:
  - duplicates may occur if same salt is used for deployment multiple times
  - so far such transactions where rejected
  - ignoring is the approach used by alpha-goerli testnet (and Devnet aims to copy it)

## Dev
- This PR doesn't contain developer-specific changes.


FabijanC:
Still pending tests.
CI is failing because we merged the PR on the example repo to early (salt test fails because the plugin functionality supporting it has not been merged to master).

dribeiro-ShardLabs (approved):
We should consider refactoring the testing in the future, `test_cli.sh` could perhaps have a more intuitive name like `init_tests.sh`, because all other scripts start with `test_*.sh`, it may be confusing.

PR #33, opened 2022-02-01
Add events support, multiple small fixes

dribeiro-ShardLabs:
Add raw_events to output ("events")
Add check for signatures
Fix tx_failure_reason object name


test/test_cli.py
```diff
@@ -63,11 +67,20 @@
 assert_equal(value, "40 60", "Checking complex input failed!")
 
 assert_salty_deploy(
```
FabijanC, line 69: I think it should be commented here that the purpose of the salty test isn't just testing salty functionality, but also testing events functionality.

FabijanC (changes requested):

FabijanC (approved):

PR #34, opened 2022-02-02
Fixed failure_reason element

dribeiro-ShardLabs:
Fixed failure_reason element

### DIFF
--- a/starknet_devnet/starknet_wrapper.py
+++ b/starknet_devnet/starknet_wrapper.py
@@ -184,13 +184,19 @@ class StarknetWrapper:
             transaction_wrapper = self.__transaction_wrappers[tx_hash_int]
             transaction = transaction_wrapper.transaction
 
+            # the transaction status object only needs 1-3 elements from the transaction_wrapper object
             ret = {
+                # "tx_status" always exists
                 "tx_status": transaction["status"]
             }
 
+            # "block_hash" will only exist after transaction enters ACCEPTED_ON_L2
             if "block_hash" in transaction:
                 ret["block_hash"] = transaction["block_hash"]
 
+            # "tx_failure_reason" will only exist if the transaction was rejected.
+            # the key in the transaction_wrapper object is "transaction_failure_reason"
+            # first it must be checked if the object contains an element with that key
             if FAILURE_REASON_KEY in transaction:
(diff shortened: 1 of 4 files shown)


starknet_devnet/starknet_wrapper.py
```diff
@@ -198,9 +198,10 @@ def get_transaction_status(self, transaction_hash: str):
             if "block_hash" in transaction:
                 ret["block_hash"] = transaction["block_hash"]
 
-            failure_key = "tx_failure_reason"
-            if failure_key in transaction:
-                ret[failure_key] = transaction[failure_key]
+            curr_failure_key = "transaction_failure_reason"
+            new_failure_key = "tx_failure_reason"
```
FabijanC, line 202: we can lose `new_failure_key`, just use `"tx_failure_reason"` directly.

starknet_devnet/starknet_wrapper.py
```diff
@@ -198,9 +198,10 @@ def get_transaction_status(self, transaction_hash: str):
             if "block_hash" in transaction:
                 ret["block_hash"] = transaction["block_hash"]
 
-            failure_key = "tx_failure_reason"
-            if failure_key in transaction:
-                ret[failure_key] = transaction[failure_key]
+            curr_failure_key = "transaction_failure_reason"
```
FabijanC, line 201: "transaction_failure_reason" can be defined in constants.py under a name like `FAILURE_REASON_KEY`.

starknet_devnet/transaction_wrapper.py
```diff
@@ -77,7 +77,7 @@ def set_failure_reason(self, error_message: str):
         assert error_message
         assert self.transaction
         assert self.receipt
-        failure_key = "tx_failure_reason"
+        failure_key = "transaction_failure_reason"
```
FabijanC, line 80: use the mentioned constant

FabijanC (changes requested):

starknet_devnet/starknet_wrapper.py
FabijanC: Also, comment that the `if key in object` is performed to check if the tx really failed and contains a failure reason.

FabijanC (approved):

PR #35, opened 2022-02-04
Test custom request

FabijanC:
## Usage
- Return `[]` instead of `{}` if invoke fails.

## Dev
- Convert test_plugin to shell script.
- Add test directory to .dockerignore.
- Use double quotes.
- Add endpoint testing (pytest):
  - To enable this, arg parsing had to be done inside the main function.
- Validate calls/transactions in separate methods.


starknet_devnet/server.py
```diff
@@ -61,6 +51,21 @@ async def add_transaction():
         **result_dict
     })
 
+def validate_transaction(data: bytes):
```
dribeiro-ShardLabs, line 54: can this function be merged with `validate_call`? they share some functionality

dribeiro-ShardLabs (approved):

PR #39, opened 2022-02-21
Get full contract

badurinantun:
## Usage
* Introduced support for the `get_full_contract` CLI command for fetching the contract definition from the provided contract address

### DIFF
--- a/test/util.py
+++ b/test/util.py
@@ -177,11 +177,10 @@ def assert_contract_definition(address, contract_path):
     ])
     contract_definition: ContractDefinition = ContractDefinition.load(json.loads(output.stdout))
 
-    with open(contract_path, encoding="utf-8") as contract_file:
-        loaded_contract = json.load(contract_file)
-        loaded_contract_definition: ContractDefinition = ContractDefinition.Schema().load(loaded_contract)
+    loaded_contract = load_json_from_path(contract_path)
+    loaded_contract_definition: ContractDefinition = ContractDefinition.load(loaded_contract)
 
-        assert_equal(contract_definition, loaded_contract_definition.remove_debug_info())
+    assert_equal(contract_definition, loaded_contract_definition.remove_debug_info())
 
 def assert_storage(address, key, expected_value):
     """Asserts the storage value stored at (address, key)."""
--- a/test/util.py
+++ b/test/util.py
@@ -9,6 +9,8 @@ import re
 import subprocess
 import time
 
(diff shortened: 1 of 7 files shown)


FabijanC (changes requested):
- ~~Can you change the PR description since you're not adding a new CLI command, but supporting it?~~
- Is just returning `contract_definition` for `get_full_contract` sufficient? Make a thorough comparison with alpha-goerli output by deploying the same contract to it (perhaps even use salt).

- If possible, include comparing alpha-goerli output with devnet output into tests. If that's not possible, consider comparing nested keys (if it makes sense).

test/util.py
```diff
@@ -161,15 +163,19 @@ def assert_contract_code(address):
...
         "--contract_address", address
     ])
-    contract_definition = json.loads(output.stdout)
-    # just checking key equality
-    assert_equal(sorted(contract_definition.keys()), ["abi", "entry_points_by_type", "program"])
+    contract_definition: ContractDefinition = ContractDefinition.load(json.loads(output.stdout))
+
+    with open(contract_path, encoding="utf-8") as contract_file:
+        loaded_contract = json.load(contract_file)
+        loaded_contract_definition: ContractDefinition = ContractDefinition.Schema().load(loaded_contract)
```
FabijanC, line 176: If `ContractDefinition` approach is kept, why not use the same approach as above (without explicitly invoking `Schema`)?

test/util.py
```diff
@@ -161,15 +163,19 @@ def assert_contract_code(address):
...
+def assert_contract_definition(address, contract_path):
     """Asserts the content of the contract definition of a contract at address."""
     output = my_run([
         "starknet", "get_full_contract",
         "--contract_address", address
     ])
-    contract_definition = json.loads(output.stdout)
-    # just checking key equality
-    assert_equal(sorted(contract_definition.keys()), ["abi", "entry_points_by_type", "program"])
+    contract_definition: ContractDefinition = ContractDefinition.load(json.loads(output.stdout))
```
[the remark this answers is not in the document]
badurinantun: I just wanted to use `remove_debug_info` method. Right now it only sets it to `null`, but perhaps it will do something else in the future

test/util.py
badurinantun: All changes were addressed in [23c20b9](https://github.com/Shard-Labs/starknet-devnet/pull/39/commits/23c20b946d8f0f13cb241b90cbb3ad0d158e27c9)

PR #40, opened 2022-02-22
Enable state dumping and loading

FabijanC:
## Usage changes
- Allow dumping devnet to a path (check the updated Readme for more detail):
  - On exit
  - On each transaction
  - On request (using e.g. curl)
- Allow loading devnet from a path

## Development changes
- Testing of this feature is added to `.circleci/config.yml`.
- Change default initial sleep time for `run_devnet_in_background`.
- Set pylint min-public-methods to 1

### DIFF
--- a/starknet_devnet/dump.py
+++ b/starknet_devnet/dump.py
@@ -1,0 +1,28 @@
+"""Dumping utilities."""
+
+import dill as pickle
+
+from .util import DumpOn
+
+class Dumper:
+    """Class for dumping objects."""
+
+    dump_path: str
+    """Where to dump."""
+
+    dump_on: DumpOn
+    """When to dump."""
+
+    def __init__(self, dumpable):
+        """Specify the `dumpable` object to be dumped."""
+        self.dumpable = dumpable
+
+    def dump(self, path: str):
+        """Dump to `path`."""
+        assert self.dump_path, "No dump_path defined"
+        with open(path, "rb") as file:
+            pickle.dump(file, self.dumpable)
+
+    def dump_if_required(self):
+        """Dump only if specified in self.dump_on."""
+        self.dump(self.dump_path)
--- a/starknet_devnet/dump.py
+++ b/starknet_devnet/dump.py
@@ -19,10 +19,13 @@ class Dumper:
 
     def dump(self, path: str):
         """Dump to `path`."""
-        assert self.dump_path, "No dump_path defined"
-        with open(path, "rb") as file:
-            pickle.dump(file, self.dumpable)
+        path = path or self.dump_path
+        assert path, "No dump_path defined"
+        print("Dumping Devnet to:", path)
+        with open(path, "wb") as file:
+            pickle.dump(self.dumpable, file)
 
     def dump_if_required(self):
         """Dump only if specified in self.dump_on."""
-        self.dump(self.dump_path)
+        if self.dump_on == DumpOn.TRANSACTION:
+            self.dump(self.dump_path)
--- a/starknet_devnet/dump.py
+++ b/starknet_devnet/dump.py
@@ -11,6 +11,7 @@ def perform_dump(dumpable, path):
     with open(path, "wb") as file:
         pickle.dump(dumpable, file)
 
+# pylint: disable=too-few-public-methods
 class Dumper:
     """Class for dumping objects."""
 
@@ -25,7 +26,7 @@ class Dumper:
         self.dump_on: DumpOn = None
         """When to dump."""
 
-    def dump(self, path: str):
+    def dump(self, path: str=None):
         """Dump to `path`."""
         path = path or self.dump_path
         assert path, "No dump_path defined"
@@ -36,8 +37,3 @@ class Dumper:
             args=[self.dumpable, path]
         ).start()
         # don't .join(), let it run in background
-
-    def dump_if_required(self):
-        """Dump only if specified in self.dump_on."""
-        if self.dump_on == DumpOn.TRANSACTION:
-            self.dump(self.dump_path)
--- a/starknet_devnet/dump.py
+++ b/starknet_devnet/dump.py
@@ -1,29 +1,41 @@
 """Dumping utilities."""
 
+import multiprocessing
+
 import dill as pickle
 
 from .util import DumpOn
 
+def perform_dump(dumpable, path):
+    """Performs the very act of dumping."""
+    with open(path, "wb") as file:
+        pickle.dump(dumpable, file)
+
 class Dumper:
     """Class for dumping objects."""
 
-    dump_path: str
-    """Where to dump."""
-
-    dump_on: DumpOn
-    """When to dump."""
-
     def __init__(self, dumpable):
         """Specify the `dumpable` object to be dumped."""
+
         self.dumpable = dumpable
+
+        self.dump_path: str = None
+        """Where to dump."""
+
+        self.dump_on: DumpOn = None
+        """When to dump."""
 
     def dump(self, path: str):
         """Dump to `path`."""
         path = path or self.dump_path
         assert path, "No dump_path defined"
         print("Dumping Devnet to:", path)
-        with open(path, "wb") as file:
-            pickle.dump(self.dumpable, file)
+
+        multiprocessing.Process(
+            target=perform_dump,
+            args=[self.dumpable, path]
+        ).start()
+        # don't .join(), let it run in background
 
     def dump_if_required(self):
         """Dump only if specified in self.dump_on."""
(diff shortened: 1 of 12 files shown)


starknet_devnet/server.py
```diff
@@ -43,7 +47,7 @@ async def add_transaction():
     else:
         abort(Response(f"Invalid tx_type: {tx_type}.", 400))
 
-    await starknet_wrapper.postman_flush()
+    dumper.dump_if_required()
```
FabijanC, line 50: @dribeiro-ShardLabs I accidentally removed the flush line, I will return it so it's not part of this PR, but just wondering - why did the postman tests pass anyway?

FabijanC: Received response in a separate channel: because of the flush in call endpoint.

test/util.py
```diff
@@ -14,13 +14,19 @@
 class ReturnCodeAssertionError(AssertionError):
     """Error to be raised when the return code of an executed process is not as expected."""
 
-def run_devnet_in_background(sleep_seconds=0):
-    """Run starknet-devnet in background. Return the process handle. Optionally sleep."""
-    command = ["poetry", "run", "starknet-devnet", "--host", HOST, "--port", PORT]
+def run_devnet_in_background(*args, sleep_seconds=3):
+    """
+    Runs starknet-devnet in background.
+    By default sleeps 1 second after spawning devnet.
```
badurinantun, line 20: Isn't the default 3 seconds now?

starknet_devnet/util.py
```diff
@@ -66,6 +69,20 @@ def fixed_length_hex(arg: int) -> str:
...
+    TRANSACTION = auto()
+
+DUMP_ON_OPTIONS = [e.name.lower() for e in DumpOn]
+DUMP_ON_OPTIONS_STRINGIFIED = ", ".join(DUMP_ON_OPTIONS)
+
+def parse_dump_on(option: str):
+    """Parse dumping frequency option."""
+    if option in DUMP_ON_OPTIONS:
+        return DumpOn[option.upper()]
+    sys.exit(f"Error: Invalid --dump-on option: {option}. Valid options: {DUMP_ON_OPTIONS_STRINGIFIED}")
```
badurinantun, line 84: Can `Enum` use lowercased string values instead of `auto()`?

starknet_devnet/util.py
```diff
@@ -99,7 +129,11 @@ def parse_args():
     #          "or network name (alpha or alpha-mainnet)",
     # )
 
-    return parser.parse_args()
+    args = parser.parse_args()
+    if args.dump_on and not args.dump_path:
```
badurinantun, line 133: Why is this check necessary?

starknet_devnet/util.py
```diff
@@ -123,3 +157,16 @@ def __init__(self):
         self.retdata = []
         self.internal_calls = []
         self.l2_to_l1_messages = []
+
+def enable_pickling():
```
badurinantun, line 161: Why is this a method and not done inside of the `StraknetContract`?

starknet_devnet/starknet_wrapper.py
```diff
@@ -55,6 +59,12 @@ def __init__(self):
         self.__l1_provider = None
         """Saves the L1 URL being used for L1 <> L2 communication."""
 
+    @classmethod
+    def load(cls, path: str) -> "StarknetWrapper":
```
badurinantun, line 63: Why is this a class method?

+ Is wrapping the return type in `"` a standard way of using the yet to be defined types?

starknet_devnet/server.py
```diff
@@ -204,19 +210,54 @@ def validate_load_messaging_contract(request_dict: dict):
...
+    dumper.dump(dump_path)
+    return Response(status=200)
+
+def dump_on_exit(_signum, _frame):
+    """Dumps on exit."""
+    dumper.dump(dumper.dump_path)
+    sys.exit(0)
+
 starknet_wrapper = StarknetWrapper()
+dumper = Dumper(starknet_wrapper)
```
badurinantun, line 231: Why are this variables declared here and not somewhere at the top of the file?

starknet_devnet/dump.py
```diff
@@ -0,0 +1,43 @@
+"""Dumping utilities."""
+
+import multiprocessing
+
+import dill as pickle
+
+from .util import DumpOn
+
+def perform_dump(dumpable, path):
```
badurinantun, line 9: Why is this not a class method?

starknet_devnet/dump.py
```diff
@@ -0,0 +1,43 @@
...
+        assert path, "No dump_path defined"
+        print("Dumping Devnet to:", path)
+
+        multiprocessing.Process(
+            target=perform_dump,
+            args=[self.dumpable, path]
+        ).start()
+        # don't .join(), let it run in background
+
+    def dump_if_required(self):
```
badurinantun, line 40: I'm not sure about this method. What if new `dump_on` trigger are added?

test/util.py
FabijanC: Right, I was experimenting with 1 second. Will change.

starknet_devnet/util.py
FabijanC: It can, but I didn't want to duplicate the naming. And I can't find a way to automatically set it to be a lowercase version of the name.

starknet_devnet/util.py
FabijanC: I think it doesn't make sense to allow user to specify **when** to dump, and not specify **where** to dump.

starknet_devnet/util.py
FabijanC: This is a monkey patch. `StarknetContract` is not our class, it's from starkware's library. By default it cannot be pickled, so this patch was necessary.

starknet_devnet/dump.py
FabijanC: I was also very unsure, but what other options are there? The logic could be inside the add_transaction endpoint.

starknet_devnet/starknet_wrapper.py
FabijanC: It doesn't use any properties of the class/object, so it doesn't make sense for it to be a regular method. It could be a standalone function, but I saw it as related to the `StarknetWrapper` class.

FabijanC: And to answer the second question, that is the standard way.

starknet_devnet/server.py
FabijanC: Back when the `main` function was mostly an `app` runner, the logic of instantiating a StarknetWrapper and parsing the args was done outside of it, so I guess this is a historic remnant. I could move them up.

starknet_devnet/dump.py
FabijanC: Not sure tbh, could be both, what do you think?

starknet_devnet/dump.py
badurinantun: Perhaps to have `dump_on` as an argument?

starknet_devnet/dump.py
badurinantun: If the dumper is the only one who'll perform the dumping we can move it to a class method

starknet_devnet/starknet_wrapper.py
badurinantun: My question is more on why is it a class method and not static method?

starknet_devnet/dump.py
FabijanC: Ended up renaming it and making it a private method: https://github.com/Shard-Labs/starknet-devnet/pull/40/commits/a9cfec85cf925ee5860c1b41918a81fdf783be76

starknet_devnet/starknet_wrapper.py
FabijanC: Made it a static method: https://github.com/Shard-Labs/starknet-devnet/pull/40/commits/a413e0a563826681652aebdd18c1c3449e75692a

test/util.py
FabijanC: Changed: https://github.com/Shard-Labs/starknet-devnet/pull/40/commits/17c17ee77c45d854787d053bace2f200de2bb31b

starknet_devnet/dump.py
FabijanC: Addressed it here: https://github.com/Shard-Labs/starknet-devnet/pull/40/commits/67fe6aa3efbd7011ce9905ea299baed0b0a4c576

badurinantun (approved):

dribeiro-ShardLabs (approved):

PR #41, opened 2022-02-25
L1 l2 improvements

dribeiro-ShardLabs:
- Removed implicit flush calls
- Improved output message from flush, including hex values instead of large integers
- Optimized postman testing


starknet_devnet/postman_wrapper.py
```diff
@@ -49,3 +50,4 @@ def load_mock_messaging_contract_in_l1(self, starknet, contract_address):
             self.mock_starknet_messaging_contract = EthContract(self.web3,address,w3_contract,abi,self.eth_account)
 
         self.postman = Postman(self.mock_starknet_messaging_contract,starknet)
+        self.l1_to_l2_message_filter = self.mock_starknet_messaging_contract.w3_contract.events.LogMessageToL2.createFilter(fromBlock="latest")
```
FabijanC, line 53: Why wasn't this needed before?

pyproject.toml
```diff
@@ -15,6 +15,7 @@ python = "^3.7"
 Flask = {extras = ["async"], version = "^2.0.2"}
 flask-cors = "^3.0.10"
 cairo-lang = "0.7.1"
+py-solc-x = "^1.1.1"
```
FabijanC, line 18: Can it be a dev dependency?

poetry.lock
```diff
@@ -33,11 +33,11 @@ frozenlist = ">=1.1.0"
 
 [[package]]
 name = "asgiref"
-version = "3.4.1"
```
FabijanC, line 36: Did all these updates come up when you installed pysolc? Or did you run an update command?

test/test_postman.py
```diff
@@ -48,26 +65,30 @@ def test_init_ganache():
...
+    set_solc_version(Version('0.6.12'))
+    install_solc_pragma("pragma solidity ^0.6.12;")
+    messaging_contract = load_file_content(STARKNET_MESSAGING_PATH)
+    l1l2_example_contract = load_file_content(L1L2_EXAMPLE_ETH_PATH)
 
-    MESSAGING_CONTRACT_ADDRESS = messaging_contract["networks"]["32"]["address"]
-    L1L2_EXAMPLE_CONTRACT_ADDRESS = l1l2_example_contract["networks"]["32"]["address"]
+    messaging_contract_interface = web3_compile(messaging_contract, os.path.abspath(SOLIDITY_CONTRACTS_PATH))
+    l1l2_example_contract_interface = web3_compile(l1l2_example_contract, os.path.abspath(SOLIDITY_CONTRACTS_PATH))
+    MESSAGING_CONTRACT_DEFINITION = web3_deploy(WEB3,messaging_contract_interface)
```
FabijanC, line 82: Why do these two variables have the _DEFINITION suffix? Are they contracts or definitions? I find it confusing that a definition holds an address.

test/web3_util.py
```diff
@@ -1,34 +1,36 @@
...
 
-    abi=json.loads(load_file_content(abi_path))["abi"]
-    contract = web3.eth.contract(address=address,abi=abi)
+def web3_transact(web3: Web3, function, contract,  *inputs):
+    """Invokes a function in a Web3 contract"""
 
     contract_function = contract.get_function_by_name(function)(*inputs)
-    tx_hash = contract_function.transact({"from": web3.eth.accounts[0], "value": 0})
+    tx_hash = contract_function.transact()
+    web3.eth.waitForTransactionReceipt(tx_hash)
```
FabijanC, line 27: What happened here, did you find a simple way of sending a transaction?

test/web3_util.py
```diff
@@ -1,34 +1,36 @@
...
 
     return tx_hash
 
-def web3_call(function, url, contract_address, abi_path, *inputs):
+def web3_call(function, contract, *inputs):
     """Invokes a function in a Web3 contract"""
```
FabijanC, line 32: Should be "calls" instead of "invokes".

test/test_postman.py
```diff
@@ -6,35 +6,52 @@
...
 import requests
 import pytest
+import os
 
-STARKNET_MESSAGING_PATH = "build/contracts/MockStarknetMessaging.json"
-L1L2_EXAMPLE_ETH_PATH = "build/contracts/L1L2Example.json"
+from solcx import install_solc, install_solc_pragma, set_solc_version
+from web3 import Web3
+from web3.contract import Contract
+from semantic_version import Version
```
FabijanC, line 22: What's this?

test/test_postman.py
```diff
@@ -6,35 +6,52 @@
...
 
-ARTIFACTS_PATH = "starknet-hardhat-example/starknet-artifacts/contracts"
+
```
FabijanC, line 24: What's with all the empty lines?

test/test_postman.py
```diff
@@ -6,35 +6,52 @@
...
+    # Setup L1 testnet
+    command = ["npx", "hardhat", "node"]
     # pylint: disable=consider-using-with
-    proc = subprocess.Popen(command, close_fds=True, stdout=subprocess.PIPE)
-    time.sleep(5)
-    atexit.register(proc.kill)
+    proc = subprocess.Popen(command,cwd="starknet-hardhat-example", close_fds=True, stdout=subprocess.PIPE)
+    atexit.register(proc.terminate)
+    time.sleep(25)
```
FabijanC, line 51: Does it really have to be that much?

test/test_postman.py
```diff
@@ -48,26 +65,30 @@ def test_init_ganache():
 
 @pytest.mark.web3_deploy
 def test_deploy_l1_contracts():
-    """Deploys Ethereum contracts in the Ganache instance, including the L1L2Example and MockStarknetMessaging contracts"""
+    """Deploys Ethereum contracts in the Hardhat testnet instance, including the L1L2Example and MockStarknetMessaging contracts"""
 
-    global MESSAGING_CONTRACT_ADDRESS # pylint: disable=global-statement
```
FabijanC, line 53: I think pylint disable statements can be applied to a whole file. But an even better solution would be this: https://stackoverflow.com/questions/44441929/how-to-share-global-variables-between-tests

test/test_postman.py
```diff
@@ -48,26 +65,30 @@ def test_init_ganache():
...
-    l1l2_example_contract = json.loads(load_file_content(L1L2_EXAMPLE_ETH_PATH))
+    install_solc(Version('0.6.12'))
+    set_solc_version(Version('0.6.12'))
+    install_solc_pragma("pragma solidity ^0.6.12;")
+    messaging_contract = load_file_content(STARKNET_MESSAGING_PATH)
+    l1l2_example_contract = load_file_content(L1L2_EXAMPLE_ETH_PATH)
 
-    MESSAGING_CONTRACT_ADDRESS = messaging_contract["networks"]["32"]["address"]
-    L1L2_EXAMPLE_CONTRACT_ADDRESS = l1l2_example_contract["networks"]["32"]["address"]
+    messaging_contract_interface = web3_compile(messaging_contract, os.path.abspath(SOLIDITY_CONTRACTS_PATH))
```
FabijanC, line 80: can't compilation artifacts be expected to exist? `npx hardhat compile` step could be added before or after `npx hardhat starknet-compile` in `setup_example.sh`: https://github.com/Shard-Labs/starknet-devnet/blob/master/scripts/setup_example.sh#L23
If that could be achieved, then the pysolc dependency (dev or not) wouldn't be needed.

FabijanC (changes requested):

pyproject.toml
FabijanC: Or is it even needed? Check this comment: https://github.com/Shard-Labs/starknet-devnet/pull/41#discussion_r814783789

starknet_devnet/postman_wrapper.py
dribeiro-ShardLabs: The way the eth filter works is it pops the results when the `get_new_entries()` function is used, and I found no other way of just retrieving the messages without them disappearing, so we need another filter to be able to return the messages to the output of `flush`

test/web3_util.py
dribeiro-ShardLabs: Yes, `WEB3.eth.default_account = WEB3.eth.accounts[0]`, by setting this in the `test_postman.py` file

starknet_devnet/postman_wrapper.py
FabijanC: Is it related to garbage collection? Like, if there is no reference to the object, it gets garbage-collected?

dribeiro-ShardLabs: I don't think so, I think it's just that `get_new_entries()`, which is also called in postman for it's filter property, simply "pops" the messages from the filter.

test/test_postman.py
dribeiro-ShardLabs: Renamed

test/test_postman.py
dribeiro-ShardLabs: You are right, that's a much simpler way. Changed do use `npx hardhat compile`, thanks

test/test_postman.py
dribeiro-ShardLabs: Changed

test/test_postman.py
dribeiro-ShardLabs: Fixed

starknet_devnet/postman_wrapper.py
FabijanC: ok, didn't realize it was used in starknet_wrapper

test/test_postman.py
dribeiro-ShardLabs: Was a dependency along with solcx to specify the version of the compiler, but it is not needed as we switched to the `npx hardhat compile` approach

pyproject.toml
dribeiro-ShardLabs: Removed according to the comment

FabijanC (approved):

PR #53, opened 2022-03-09
Improve Docker container size

LucasLvy:
Initial size : 634MB
Improved size : 225MB

### DIFF
--- a/README.md
+++ b/README.md
@@ -65,12 +65,12 @@ docker pull shardlabs/starknet-devnet
 
 The server inside the container listens to the port 5000, which you need to publish to a desired `<PORT>` on your host machine:
 ```text
-docker run -it -p [HOST:]<PORT>:5000 shardlabs/starknet-devnet
+docker run -p [HOST:]<PORT>:5000 shardlabs/starknet-devnet
 ```
 
 E.g. if you want to use your host machine's `127.0.0.1:5000`, you need to run:
 ```text
-docker run -it -p 127.0.0.1:5000:5000 shardlabs/starknet-devnet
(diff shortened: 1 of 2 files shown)


FabijanC (commented):
Thank you very much for this PR, looks great. Basically the biggest game changer was having a separate builder? Are there also any performance benefits? Glad to receive support from someone with more Docker experience. I'll try running locally and then see about final approval (probably tomorrow).

LucasLvy:
> Basically the biggest game changer was having a separate builder? 

Yes so we can get rid of the apk packages only needed for the installation of the python packages. The python alpine container also helps a lot.

> Are there also any performance benefits? 

Comparing with the devnet without docker no there's no performance gain, it's really similar the transaction speed processing.
Comparing to the old docker container I haven't tried that much so I can't say.

README.md
```diff
@@ -157,23 +157,19 @@ This example:
 
 If there is `dump.pkl` inside `/actual/dumpdir`, you can load it with:
 ```
-docker run -it \
+docker run \
   -p 127.0.0.1:5000:5000 \
   --mount type=bind,source=/actual/dumpdir,target=/dumpdir \
   shardlabs/starknet-devnet \
-  poetry run starknet-devnet \
-  --host 0.0.0.0 --port 5000 \
```
FabijanC, line 165: Since this is changed, can you also change line 156 so that it doesn't mention the second sentence:
> However, leave the `--host 0.0.0.0` part as it is.

FabijanC (changes requested):

PR #73, opened 2022-04-11
Test account fees

badurinantun:


starknet_devnet/starknet_wrapper.py
```diff
@@ -530,12 +533,30 @@ def get_state_update(self, block_hash=None, block_number=None):
...
-    async def calculate_actual_fee(self, transaction: InvokeFunction):
+    async def calculate_actual_fee(self, external_tx: InvokeFunction):
         """Calculates actual fee"""
         state = await self.__get_state()
-        internal_tx = InternalInvokeFunction.from_external(transaction, state.general_config)
+        internal_tx = InternalInvokeFunction.create(
+            contract_address=external_tx.contract_address,
+            entry_point_selector=external_tx.entry_point_selector,
+            max_fee=external_tx.max_fee,
+            entry_point_type=EntryPointType.EXTERNAL,
```
FabijanC, line 543: Is it safe to always set to `EXTERNAL`?

starknet_devnet/starknet_wrapper.py
```diff
@@ -530,12 +533,30 @@ def get_state_update(self, block_hash=None, block_number=None):
...
+            calldata=external_tx.calldata,
+            signature=external_tx.signature,
+            nonce=None,
```
FabijanC, line 546: Is there no way to pass some actual `nonce` from `external_tx`?

starknet_devnet/starknet_wrapper.py
badurinantun: This is what was set in the `from_external` before, so no changes were made regarding that. Same as `nonce`

FabijanC (approved):

PR #81, opened 2022-04-12
JSON Error response

Nathan-SL:
## Usage related changes

- An error will be displayed in `JSON` format rather than HTML

Exceptions will be picked up by the new handler decorated by `@app.errorhandler()` inside `server.py`
- Use `raise StarknetDevnetException()` for throwing errors.
- Exceptions of type `StarkException` and `StarknetDevnetException` will also be handled.
- Additionally, using `try/except` block should be avoided on calls to `startknet_wrapper` like below.
```python
result_dict = await state.starknet_wrapper.call(call_specifications)
```

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


starknet_devnet/util.py
```diff
@@ -241,3 +242,21 @@ def generate_state_update(previous_state: CarriedState, current_state: CarriedSt
             "storage_diffs": storage_diffs
         }
     }
+
+class JsonErrorHandler:
```
badurinantun, line 246: Could we leverage use of [Flask.errorhandler](https://flask.palletsprojects.com/en/2.1.x/api/#flask.Flask.errorhandler)? We could catch either status or `StarknetDevnetException` and return JSON there?

test/test_error_responses.py
```diff
@@ -0,0 +1,200 @@
+"""
```
FabijanC, line 1: I wouldn't have all the error tests in a single file. I think they should be put into their corresponding test files; e.g. tests testing dumping errors should go to the test_dump.py file. Tell me if you don't think some tests don't have a corresponding test file.

starknet_devnet/util.py
```diff
@@ -241,3 +242,21 @@ def generate_state_update(previous_state: CarriedState, current_state: CarriedSt
...
+    """
+    Error handler that outputs errors in JSON format.
+    """
+
+    def __init__(self, error_message, error_code):
+        self.error_code = error_code
+        self.error_message = error_message
+
+    def handle(self):
```
FabijanC, line 255: Do you think there is benefit from having this as a class rather than a simple function that accepts the error_message and error_code?

FabijanC (changes requested):

starknet_devnet/util.py
Nathan-SL: So, now, the whole `JsonErrorResponse` class is now removed. https://github.com/Shard-Labs/starknet-devnet/pull/81/commits/ac86499da19eb8d3d21b35c52cd0980efd1be8fb

test/test_error_responses.py
Nathan-SL: https://github.com/Shard-Labs/starknet-devnet/pull/81/commits/c83f3afe5b84a5da50fd55c6cd98b9fa28006090

starknet_devnet/util.py
Nathan-SL: https://github.com/Shard-Labs/starknet-devnet/pull/81/commits/8ab904c4af7208bf5e4441facd4861f5d74ab257

FabijanC (approved):

Issue #100, opened 2022-05-16
Max_fee missing from transaction


cptartur:
It seems like `max_fee` field is missing from transactions returned by `feeder_gateway/getTransaction`. E.g. see this random [transaction from voyager](https://goerli.voyager.online/tx/0x23eaff194a2b4138372ae78f62591c252a603595092ed15a7aba28e0b34aae2). If you [query it from feeder gateway directly](https://alpha4.starknet.io/feeder_gateway/get_transaction?transactionHash=0x23eaff194a2b4138372ae78f62591c252a603595092ed15a7aba28e0b34aae2) you get a response containing a `max_fee` field (I believe this is different from `actual_fee` from transaction receipt. 

This field is missing from transactions returned by devnet: 
1. Deploy any contract to devnet
2. Make a invoke call to that contract
3. GET `devnet_address/feeder_gateway/get_transaction?transactionHash=hash_of_tx`, field `max_fee` will be missing from the response.

PR #101, opened 2022-05-16
Feat/rpc support

cptartur:
## Usage related changes

- Added a /rpc endpoint that allows for making RPC calls according to the [starknet rpc specs](https://github.com/starkware-libs/starknet-specs/blob/ec01ba5fd12d4a51a9202146a2d6247eebc08644/api/starknet_api_openrpc.json) - v0.8.0

None

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

## Additional details

I didn't want to modify the internals of devnet much while implementing this feature and relied on the methods already present in `starknet_wrapper`. Because of that I had to do some "hacks" in few methods. If you think it would be better to instead make modifications/add methods to wrapper and use them instead I'm open for any suggestions.

Also lint checks will fail with this PR because of two `FIXME` in tests and rpc, they will have to be removed later (especially after #100 is closed).


test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+    invoke_tx = json.loads(INVOKE_CONTENT)
+    invoke_tx["calldata"] = ["0"]
+    resp = send_transaction(invoke_tx)
+    invoke_info = json.loads(resp.data.decode("utf-8"))
+    return invoke_info
+
+
+def pad_zero(felt: str) -> str:
+    """
+    Convert felt with format `0xValue` to format `0x0Value`
```
FabijanC, line 77: Is always adding a leading zero the right solution? Shouldn't it not add the zero only if the length is not as required? I.e., it depends on how big the number is. (That's why sometimes converting the hex to int is used for testing).

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+def test_get_block_by_hash(deploy_info):
+    """
+    Get block by hash
+    """
+    block_hash = gateway_call("get_block", blockNumber=0)["block_hash"]
+
+    resp = rpc_call(
+        "starknet_getBlockByHash", params={"block_hash": block_hash}
+    )
+    block2 = resp["result"]
```
FabijanC, line 127: Why block2?

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+    expected = {
+        "block_hash": block_hash,
+        "parent_hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+        "block_number": 0,
+        "status": "ACCEPTED_ON_L2",
+        "sequencer": "0x0000000000000000000000000000000000000000",
+        "old_root": "0x0",
+    }
+
+    for key, value in expected.items():
```
FabijanC, line 137: Can't you just compare them with `==`? Like `assert block == expected`? It would be simpler and would make sure there aren't extra properties.

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+
+def pad_zero(felt: str) -> str:
+    """
+    Convert felt with format `0xValue` to format `0x0Value`
+    """
+    felt = felt.lstrip("0x")
+    return "0x0" + felt
+
+
+# pylint: disable=unused-argument
```
FabijanC, line 83: Why suppresion? Can't you just not have the deploy_info parameter?

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+        }
+    )
+
+    assert ex["error"] == {
+        "code": 20,
+        "message": "Contract not found"
+    }
+
+
+# FIXME internal workings of get_storage_at would have to be changed for this to work
```
FabijanC, line 200: How would you solve this by changing internal logic of feeder_gateway's get_storage_at?

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+    assert storage == "0x0"
+
+
+def test_get_storage_at_raises_on_incorrect_contract(deploy_info):
+    """
+    Get storage at incorrect contract
+    """
+    block = gateway_call("get_block", blockNumber=0)
+
+    key: str = str(hex(get_storage_var_address("balance")))
```
FabijanC, line 183: No need for applying `str` on `hex` as `hex` already returns a string. In other places as well.

FabijanC (changes requested):
- #100 has been merged. Does this change what you wrote in the PR description?
- What about the rebasing to the target branch? I see it's marked as TODO in your PR description. Also, just before merging your PR, I will probably change the target branch to something other than master (but on the same commit), to see how testing goes.
- In my comments I may identified at least some of the hacks you are mentioning. Did I identify all of them?
- You could consider adding an e2e test with the run_devnet_in_background annotation (see other tests for example)
- Endpoints not tested:
  - getStateUpdateByHash
  - blockNumber
- Regarding RPC endpoints not implemented due to Devnet limitations:
  - Not saying you should do it, but have you considered making them return a Not Implemented response?
  - Couldn't chainId endpoint be implemented?

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,420 @@
...
+    """
+    try:
+        result = state.starknet_wrapper.blocks.get_state_update(block_hash=block_hash)
+    except StarknetDevnetException as ex:
+        raise RpcError(code=24, message="Invalid block hash") from ex
+
+    return rpc_state_update(result)
+
+
+async def get_storage_at(contract_address: str, key: str, block_hash: str) -> str:
```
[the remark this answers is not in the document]
cptartur: RPC takes a additional `block_hash` parameter for the `get_storage_at`, but I agree with with you that current implementation doesn't make sense. Perhaps we could do it like `call` where it only accepts `latest` as a block_hash. What do you think?

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,420 @@
...
+        if "While handling calldata" in ex.message:
+            raise RpcError(code=22, message="Invalid call data") from ex
+        raise RpcError(code=-1, message=ex.message) from ex
+
+
+async def get_block_number() -> int:
+    """
+    Get current number of blocks
+    """
+    result = state.starknet_wrapper.blocks.get_number_of_blocks() - 1
```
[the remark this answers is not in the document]
cptartur: Docstring is wrong, it's supposed to return latests accepted block number.

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,420 @@
...
+            in block["transactions"]
+        ]
+        transactions = [{**transaction, **receipt} for transaction, receipt in zip(transactions, receipts)]
+
+    return {
+        "block_hash": block["block_hash"],
+        "parent_hash": block["parent_block_hash"] or "0x0",
+        "block_number": block["block_number"],
+        "status": block["status"],
+        "sequencer": "0x0000000000000000000000000000000000000000",
```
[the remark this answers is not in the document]
cptartur: I didn't realise `general_config.sequencer_address` existed so I hardcoded this value. Will change that

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,420 @@
...
+        "accepted_time": 0,  # hardcoded value as state_update dict does not contain timestamp
+        "state_diff": {
+            "deployed_contracts": [
+                rpc_state_diff_contract(contract)
+                for contract
+                in state_update["state_diff"]["deployed_contracts"]
+            ],
+            "storage_diffs": [
+                {
+                    **{
```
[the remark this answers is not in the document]
cptartur: This combines dict i defined with `storage` which is also a dict. I think it's more concise this way.

test/test_rpc_endpoints.py
cptartur: That one I'm unsure about. In theory, the RPC has [it's own definition of felt](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_api_openrpc.json#L819) so ideally we should always be using `0x0` padded strings. 

But I don't have a strong opinion on that, I wrote this function because for some reason I was getting `0x0` strings from rpc (which I wanted), and `0x` from gateway, and for testing sake they needed to be uniform format.

test/test_rpc_endpoints.py
cptartur: Maybe `deploy_info` is not a good name, or a different fixture structure should be used instead?

Deploy_info fixture makes sure that there's always a block with some transaction before any of these tests is run. The problem was when i was running tests like `get_test_storage_at` on its own, it would fail because it tried to do `gateway_call("get_block", blockNumber=0)` even though there was no blocks.  

Perhaps this fixture should be changed to `@pytest.fixture(autouse=True)`?

test/test_rpc_endpoints.py
cptartur: My mistake, leftover from the previous version of this test.

test/test_rpc_endpoints.py
cptartur: Agreed, this will be changed

test/test_rpc_endpoints.py
cptartur: Oops

test/test_rpc_endpoints.py
cptartur: So afaik RPCs `get_storage_at` allows for this additional `block_hash` parameter so you can access storage "from the past". For this to work in devnet, internal get_storage_at would also have to accept parameter like that. 

Interestingly, I think that mainnet/testnet gateway also has `block_hash` and `block_number` parameters, but I'm unsure how they influence the returned value.

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,420 @@
...
+    """
+    Get the details of a transaction by a given block hash and index
+    """
+    try:
+        block = state.starknet_wrapper.blocks.get_by_hash(block_hash=block_hash)
+    except StarknetDevnetException as ex:
+        raise RpcError(code=24, message="Invalid block hash") from ex
+
+    try:
+        transaction_hash = block["transactions"][index]["transaction_hash"]
```
cptartur, line 115: I would consider this a hack, also see method below.

cptartur:
> * #100 has been merged. Does this change what you wrote in the PR description?
> * What about the rebasing to the target branch? I see it's marked as TODO in your PR description. Also, just before merging your PR, I will probably change the target branch to something other than master (but on the same commit), to see how testing goes.

I'm currently updating PR because it is missing changes introduced by wrapper refactor.

> * In my comments I may identified at least some of the hacks you are mentioning. Did I identify all of them?

Most of them, I would also consider what i did in `get_transaction_by_block_hash_and_index`, and `get_transaction_by_block_number_and_index` a hacks.

> * You could consider adding an e2e test with the run_devnet_in_background annotation (see other tests for example)

I considered that but I though these endpoint tests covered most of the use cases. I may try adding a test based on "general_workflow" tests but I'm not sure how that will go.

> * Endpoints not tested:
>   
>   * getStateUpdateByHash
>   * blockNumber

Those will need to be added

> * Regarding RPC endpoints not implemented due to Devnet limitations:
>   
>   * Not saying you should do it, but have you considered making them return a Not Implemented response?

I think we could do something like that. Maybe throw a custom RPC error so they have consistent returns with other methods?

>   * Couldn't chainId endpoint be implemented?

I wasn't sure what value to return there. I see pathfinder now added this parameter to their full node, so I'll look how they do this endpoint and maybe implement it.

test/test_rpc_endpoints.py
FabijanC: I'm referring to the pylint suppresion. It can be avoided by simply making function `test_get_block_by_number` not accept any parameters (since it doesn't use any, not even `deploy_info`).

test/test_rpc_endpoints.py
FabijanC: Alright, it can stay like this, doesn't look wrong.

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,420 @@
...
+    # in devnet is more complicated if possible at all
+    if block_hash != "latest":
+        # By RPC here we should return `24 invalid block hash` but in this case I believe it's more
+        # descriptive to the user to use a custom error
+        raise RpcError(code=-1, message="Calls with block_hash != 'latest' are not supported currently.")
+
+    try:
+        return await state.starknet_wrapper.call(transaction=make_invoke_function(request_body))
+    except StarknetDevnetException as ex:
+        if "No contract at the provided address" in ex.message:
```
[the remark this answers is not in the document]
cptartur: I think `"No contract at the provided address" in ex.message` could probably be replaced with something like get_transaction at address and verifying its return. 

But the other exceptions (entrypoint, handling calldata...) would require this derived exceptions as you suggested. I don't know how much implementing that would interfere with how gateway/feeder_gateway currently works though.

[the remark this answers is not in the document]
cptartur: `"No contract at the provided address" in ex.message:` can probably be replaced by contracts.is_deployed check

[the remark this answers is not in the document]
cptartur: This has been replaced with contracts.is_deployed

test/test_rpc_endpoints.py
cptartur: Most of the tests now use`assert block == expected`

test/test_rpc_endpoints.py
cptartur: The way get_storage_at works now is it only accepts `blockHash="latest"` which is the same as `call`. I don't think we can realistically checking storage in the past though I'm open for suggestions.

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,641 @@
...
+    Get contract code
+    """
+    contract_address: str = deploy_info["address"]
+
+    resp = rpc_call(
+        "starknet_getCode", params={"contract_address": contract_address}
+    )
+    code = resp["result"]
+
+    assert "bytecode" in code
```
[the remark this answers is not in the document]
cptartur: I'm comparing this with response from gateway now. I've also considered having bytecode and abi in separate file and comparing it to that, but I don't know if this is necessary?

cptartur:
* I've implemented improvements based on most of the comments.
* I've also introduced better typing to the rpc methods: there're now using TypedDicts. I think it will allow for less mistakes in the future.
* ChainId endpoint has been implemented, as well as the protocol version.
* I've added missing test, not e2e test though, do you think we should have them for rpc endpoint?
* Not implemented endpoints now raise `NotImplementedError()`

I think we're getting closer to being merge-ready but we're still not quite there yet: there are 3 FIXMEs in the code that will need to be discussed, I will add comments for them. I've marked as resolved some of the most obvious discussions, but there are still some that need to be addressed. 

I've also uncovered something that I think is a minor bug in devnet that is causing one of my tests to fail: I will open an issue with steps to reproduce soon.

Also, unfortunately RPC changes faster than I implement new stuff 😅 and 0.9.0 of the rpc released yesterday (and i see 0.10.0 is already in the works). Most of the changes brought by this version are related to starknet 0.9.0 which is not out as of writing this message, so it shouldn't block a merge this PR, but it might anytime soon (not saying we should hurry though, as these changes will need to be supported anyway).

tl;dr i did some fixes, but we're not 100% done yet

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,681 @@
...
+                "contract_hash": contract["contract_hash"]
+            }
+            _contracts.append(diff)
+        return _contracts
+
+    rpc_state: RpcStateUpdate = {
+        "block_hash": state_update["block_hash"],
+        "new_root": state_update["new_root"],
+        "old_root": state_update["old_root"],
+        "accepted_time": 0, # FIXME remove hardcoded value
```
cptartur, line 420: Gateway state update doesn't have an accepted time at all so I'm not sure how to replace that

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,803 @@
...
+    assert "old_root" in state_update
+    assert isinstance(state_update["old_root"], str)
+    assert "accepted_time" in state_update
+    assert isinstance(state_update["accepted_time"], int)
+    assert state_update["state_diff"] == {
+        "storage_diffs": [],
+        "contracts": [
+            {
+                "address": contract_address,
+                "contract_hash": "06f8d704f5a8f6bcb53a1c87c6f8d466ab0aaa5cf962084ac03a2145aac2d823",  # FIXME this is hardcoded but I don't know how this contract
```
cptartur, line 244: I've got this value by looking at what gateway state update is returning now. I'd like to calculate it correctly, but I don't quite understand how contracts are hashed for the state update purpose

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,803 @@
...
+    assert "new_root" in state_update
+    assert isinstance(state_update["new_root"], str)
+    assert "old_root" in state_update
+    assert isinstance(state_update["old_root"], str)
+    assert "accepted_time" in state_update
+    assert isinstance(state_update["accepted_time"], int)
+    assert state_update["state_diff"] == {
+        "storage_diffs": [
+            {
+                # FIXME this will fail because of bug in devnet
```
cptartur, line 266: I think storage diffs should show that there was a change when balance increased from 0 to 10, but it doesn't. I will be opening a proper issue soon.

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,681 @@
...
+    return transaction
+
+
+def rpc_deploy_transaction(transaction: DeploySpecificInfo) -> RpcTransaction:
+    """
+    Convert gateway deploy transaction to rpc format
+    """
+    def calldata() -> Optional[List[str]]:
+        # pylint: disable=no-member
+        _calldata = transaction.constructor_calldata
```
cptartur, line 481: Deploy transactions don't have `calldata` in gateway but they have `constructor_calldata`. For the rpc purpose I've decided to treat it as calldata, but for example that's no the case in the pathfinder fullnode implementation. 

But that's something RPC standard doesn't really explain.

starknet_devnet/blueprints/rpc.py
FabijanC: Is taking the block timestamp an option?

test/test_rpc_endpoints.py
FabijanC: Perhaps the compute_contract_hash from starkware.starknet.core.os.contract_hash?

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,681 @@
...
+    accepted_time: int
+    transactions: List[str] | List[dict]
+
+
+async def rpc_block(block: StarknetBlock, requested_scope: Optional[str] = "TXN_HASH") -> RpcBlock:
+    """
+    Convert gateway block to rpc block
+    """
+    async def transactions() -> List[RpcTransaction]:
+        # pylint: disable=no-member
```
FabijanC, line 315: Why this?

starknet_devnet/blueprints/rpc.py
FabijanC: I don't understand; also, why disabling the linting here?

starknet_devnet/blueprints/rpc.py
```diff
@@ -0,0 +1,681 @@
...
+    """
+    chain: int = DEFAULT_GENERAL_CONFIG.chain_id.value
+    return hex(chain)
+
+
+async def pending_transactions():
+    """
+    Returns the transactions in the transaction pool, recognized by this sequencer
+    """
+    raise NotImplementedError()
```
FabijanC, line 256: When I mentioned Not Implemented, I meant the 501 Not Implemented HTTP response

starknet_devnet/blueprints/rpc.py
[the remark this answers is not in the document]
FabijanC: if you think that's better, you can do it with only accepting latest. Otherwise keep it like this.

starknet_devnet/blueprints/rpc.py
cptartur: I think we could do that. Since gateway state_update doesn't have a accepted time, this would require additional call to `get_block`, but that's probably not a big deal.

test/test_rpc_endpoints.py
cptartur: I'll try that

starknet_devnet/blueprints/rpc.py
cptartur: Oh, I've misunderstood then. Perhaps I can add additional `except NotImplementedError` to base rpc route and then return a 501 http response. 

Another option would be to create return RPC error like all other errors with custom code (like `-2`) and a message.

Which one you think is better?

test/test_rpc_endpoints.py
cptartur: This will be updated

starknet_devnet/blueprints/rpc.py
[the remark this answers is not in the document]
cptartur: I think it's better to only accept `latest`. I'm unsure on how gateway's get_storage_at works with block_hash, and I'm assuming RPCs should return storage as it was in the pass block, referenced by given block_hash.

starknet_devnet/blueprints/rpc.py
cptartur: I meant that DeployTransactions have `constructor_calldata` instead of `calldata` like InvokeTransactions. I've decided that in RPC I'll treat `constructor_calldata` == `calldata` because it sounds logical to me, but the RPC specs aren't clear on that.

About the disabling lint, now I'm not sure, I'll check if it passes lints without disable and reply here/remove that.

starknet_devnet/blueprints/rpc.py
cptartur: Same as above

cptartur:
> * Instead of importing DEFAULT_GENERAL_CONFIG, do you think it would be better to use the `general_config` property of `state.starknet_wrapper.get_state()`?

I didn't realise this was exposed though starknet_wrapper. It's definitely better to used that

> * You don't have to introduce an e2e test.

👍  I think they aren't needed that much here since wrapper is tested anyway through e2e tests that already exists, and RPC is just another layer on that

starknet_devnet/blueprints/rpc.py
FabijanC: I like the approach with a custom code (-2).

starknet_devnet/blueprints/rpc.py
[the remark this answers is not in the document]
FabijanC: ok then, only accepting `latest`

starknet_devnet/blueprints/rpc.py
cptartur: Added custom error with -2

starknet_devnet/blueprints/rpc.py
cptartur: I've tested this and for some reason pylint doesn't play nice with these nested functions and thinks transaction is a Dict which is not correct. Same in the other places.

test/test_rpc_endpoints.py
cptartur: Hash is now calculated correctly

starknet_devnet/blueprints/rpc.py
cptartur: Done

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,820 @@
...
+import typing
+
+import pytest
+from starkware.starknet.public.abi import get_storage_var_address, get_selector_from_name
+from starkware.starknet.services.api.contract_class import ContractClass
+from starkware.starknet.core.os.class_hash import compute_class_hash
+from starkware.starknet.services.api.gateway.transaction import Transaction, Deploy
+
+from starknet_devnet.server import app
+from starknet_devnet.general_config import DEFAULT_GENERAL_CONFIG
```
FabijanC, line 16: The DEFAULT_GENERAL_CONFIG is still used in some places, can it be replaced with state.general_config?

test/test_rpc_endpoints.py
```diff
@@ -0,0 +1,820 @@
...
+    """
+    Make a call to the gateway
+    """
+    resp = app.test_client().get(
+        f"/feeder_gateway/{method}?{'&'.join(f'{key}={value}&' for key, value in kwargs.items())}"
+    )
+    return json.loads(resp.data.decode("utf-8"))
+
+
+@pytest.fixture(name="contract_definition")
```
FabijanC, line 56: Change naming to contract_class

FabijanC (changes requested):

test/test_rpc_endpoints.py
FabijanC: After discussing this - it makes sense to use DEFAULT_GENERAL_CONFIG directly in tests.

FabijanC (approved):

FabijanC (approved):

PR #104, opened 2022-05-18
Use types

badurinantun:
## Usage related changes

-

-

## Checklist:

- [ ] Performed a self-review of the code
- [ ] Rebased to the base branch
- [ ] Documented the changes
- [ ] Updated the tests


starknet_devnet/blocks.py
```diff
@@ -89,52 +86,53 @@ def get_state_update(self, block_hash=None, block_number=None):
...
-            signature = [int(sig_part, 16) for sig_part in tx_wrapper.transaction["transaction"]["signature"]]
+        signature = transaction.get_signature()
 
-        parent_block_hash = self.__get_last_block()["block_hash"] if block_number else fixed_length_hex(0)
-        sequencer_address = state.general_config.sequencer_address
-        gas_price = state.general_config.min_gas_price
+        if block_number == 0:
+            parent_block_hash = 0
+        else:
+            last_block: StarknetBlock = self.__get_last_block()
```
FabijanC, line 104: Why declare last_block type explicitly, can't the return type of `__get_last_block` be used?

starknet_devnet/origin.py
```diff
@@ -63,21 +71,28 @@ class NullOrigin(Origin):
...
-            "transaction_hash": transaction_hash,
-            "events": []
-        }
+    def get_transaction(self, transaction_hash: str) -> TransactionInfo:
+        return TransactionInfo.create(
+            status=TransactionStatus.NOT_RECEIVED,
+        )
+
+    def get_transaction_receipt(self, transaction_hash: str) -> TransactionReceipt:
+        return TransactionReceipt(
```
FabijanC, line 83: Would it be faster to have this object created in advance?

starknet_devnet/transactions.py
```diff
@@ -1,11 +1,150 @@
...
+            error_message=error_message
+        )
+
+    def get_signature(self) -> List[int]:
+        """Returns the signature"""
+        return self.internal_tx.signature if hasattr(self.internal_tx, "signature") else []
+
+    def get_tx_info(self) -> TransactionInfo:
+        """Returns the transaction info"""
+        return TransactionInfo.create(
```
FabijanC, line 102: If performance is an issue, maybe have this cached, `if not self.tx_info: self.tx_info = ...; ... return self.tx_info`. Same for other returned objects (TransactionTrace, TransactionExecution).

starknet_devnet/transactions.py
```diff
@@ -1,11 +1,150 @@
...
+        """Returns the block number"""
+        return self.block.block_number if self.block else None
+
+    def set_block(self, block: StarknetBlock):
+        """Sets the block hash and number of the transaction"""
+        self.block = block
+
+    def set_failure_reason(self, error_message: str):
+        """Sets the failure reason of the transaction"""
+        # set tx_id?
```
FabijanC, line 90: TODO?

test/test_deploy.py
```diff
@@ -0,0 +1,80 @@
+"""Test devnet contract deployment"""
```
FabijanC, line 1: I see that these are unit tests, but would it make sense to move here the salty_deploy and failing_deploy tests from test_general_workflow.py?

FabijanC (changes requested):
What's the conclusion of time performance impact of the made changes?

starknet_devnet/transactions.py
badurinantun: Perfomance is not the issue, it was because deployment tx hash was calculated twice.

We don't want to cache receipt and info as they change depending on the block information. Adding updates and caching would introduce unnecessary overhead IMHO and would proved minimal benefit.

starknet_devnet/transactions.py
badurinantun: There is no need to change this as `TransactionFailureReason` doesn't have that property

starknet_devnet/origin.py
badurinantun: It would be slower for users that never call `get_transaction_receipt`

starknet_devnet/transactions.py
FabijanC: Ok, and now the double calculation issue is resolved?

badurinantun: Correct

starknet_devnet/blocks.py
badurinantun: [f7f7e48](https://github.com/Shard-Labs/starknet-devnet/pull/104/commits/f7f7e4821c8bb000b32f5d7d8904253b1ecfade2)

PR #105, opened 2022-05-19
Fix --lite-mode-block-hash behavior

FabijanC:
## Usage related changes

- Allow timestamp manipulation in lite-mode
- Document that state updates aren't supported with --lite-mode-block-hash

- In `__update_state` method, the whole body was executed if not running in lite-mode-block-hash:
  - Now the only part dependent on lite-mode-block-hash is state_preservation
  - Rationale: the preservation part of the method takes around 89% of the whole method execution time
- Add compatibility test of lite-mode and timestamp manipulation

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -251,9 +251,9 @@ Consider passing these CLI flags on Devnet startup:
 - `--lite-mode` enables all of the optimizations described below (same as using all of the flags below)
 - `--lite-mode-deploy-hash`
   - disables the calculation of transaction hash for deploy transactions
-  - disables get_state_update functionality
 - `--lite-mode-block-hash`
   - disables the calculation of block hash
+  - disables get_state_update functionality
 
 ## Restart
 
--- a/README.md
+++ b/README.md
@@ -249,8 +249,11 @@ To improve Devnet performance, instead of calculating the actual hash of deploym
 Consider passing these CLI flags on Devnet startup:
 
 - `--lite-mode` enables all of the optimizations described below (same as using all of the flags below)
-- `--lite-mode-deploy-hash` disables the calculation of transaction hash for deploy transactions
-- `--lite-mode-block-hash` disables the calculation of block hash
+- `--lite-mode-deploy-hash`
+  - disables the calculation of transaction hash for deploy transactions
+  - disables get_state_update functionality
+- `--lite-mode-block-hash`
+  - disables the calculation of block hash
 
 ## Restart
 
--- a/starknet_devnet/starknet_wrapper.py
+++ b/starknet_devnet/starknet_wrapper.py
@@ -87,26 +87,27 @@ class StarknetWrapper:
         return starknet.state
 
     async def __update_state(self):
+        previous_state = self.__current_carried_state
+        assert previous_state is not None
+        current_carried_state = (await self.get_state()).state
+        state = await self.get_state()
+
+        current_carried_state.block_info = self.block_info_generator.next_block(
+            block_info=current_carried_state.block_info,
+            general_config=state.general_config
+        )
+
+        updated_shared_state = await current_carried_state.shared_state.apply_state_updates(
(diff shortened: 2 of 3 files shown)


starknet_devnet/starknet_wrapper.py
```diff
@@ -87,26 +87,27 @@ async def get_state(self):
...
-            await self.__preserve_current_state(self.__starknet.state.state)
+        updated_shared_state = await current_carried_state.shared_state.apply_state_updates(
+            ffc=current_carried_state.ffc,
+            previous_carried_state=previous_state,
+            current_carried_state=current_carried_state
+        )
+
+        if not self.config.lite_mode_block_hash:
+            state.state.shared_state = updated_shared_state
+            await self.__preserve_current_state(state.state)
```
badurinantun, line 108: Why this shouldn't be executed in the lite mode?

FabijanC: It's the heaviest part of the method, as noted in the PR description. In lite mode, without it, timestamp manipulation is still possible, but getting real block hash and state updates is not.

badurinantun: What do you mean by heaviest? Why is it needed for the regular workflow and not lite?

FabijanC: Heaviest time-wise (as noted, takes around 89% of the whole method time). So with only skipping that in lite-mode, we still get the time benefit. And in regular mode it's needed for state update calculation and block state_root calculation.

FabijanC: Maybe also `return generate_state_update` part can be indented (put under the if part, as it currently is on the master branch)

badurinantun: This is not evident from the code so we should add a comment which explains it.

We should refactor this in the future so that state update and block generation are clearer.

README.md
```diff
@@ -249,8 +249,11 @@ To improve Devnet performance, instead of calculating the actual hash of deploym
 Consider passing these CLI flags on Devnet startup:
 
 - `--lite-mode` enables all of the optimizations described below (same as using all of the flags below)
-- `--lite-mode-deploy-hash` disables the calculation of transaction hash for deploy transactions
-- `--lite-mode-block-hash` disables the calculation of block hash
+- `--lite-mode-deploy-hash`
+  - disables the calculation of transaction hash for deploy transactions
+- `--lite-mode-block-hash`
+  - disables the calculation of block hash
+  - disables get_state_update functionality
```
badurinantun, line 256: Isn't this enabled now? `__update_state` returns the state_update and it's saved on the block?

badurinantun (changes requested):

README.md
FabijanC: When you say "now", do you mean currently on master or with the changes introduced by this PR?

starknet_devnet/starknet_wrapper.py
FabijanC: https://github.com/Shard-Labs/starknet-devnet/pull/105/commits/b643769c7ebe4bf5639343625b7b2494722c55bf

README.md
FabijanC: Actually neither the behavior on master nor behavior introduced with this PR are good. What I meant when I wrote `disables get_state_update functionality` was that it doesn't work as expected. What would be better is to log a warning or return an error from the endpoint.

FabijanC: Commit https://github.com/Shard-Labs/starknet-devnet/pull/105/commits/bf050eb33688921d0e59598f98964e9b272c50ae restores the old behavior of returning `None` in case of using lite mode. To make it clear that state updates aren't supported (in the sense that the response is not what the user would expect), it's either this (which results in state update being `null`) or returning an error.

PR #108, opened 2022-05-26
Bump Python max version

ca11ab1e:
I guess upcoming Python 3.9 minor releases will not break anything,
so it should be safe to simply state we do support Python until 3.10.

To be more confident, I updated the CI configuration to test min/max supported
Python versions.

## Checklist:

- [ ] Documented the changes
- [ ] Updated the tests

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,10 +8,11 @@ aliases:
         only:
           - master
     requires:
-      - test
+      - test-py-min
+      - test-py-max
 
 jobs:
-  test:
+  test-py-min: &test-template
     docker:
       - image: cimg/python:3.7.12-node
     resource_class: medium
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -42,13 +43,20 @@ jobs:
             poetry run pytest -s --junitxml=test-results/junit.xml -vv $TESTFILES
       - store_test_results:
           path: test-results
+
+  test-py-max:
+    <<: *test-template
+    docker:
+      - image: cimg/python:3.9.12-node
```
FabijanC, line 50: Wouldn't it be better to have it as `cimg/python:3.9-node`? That way, the latest patch version of 3.9 would be used

FabijanC (changes requested):

.circleci/config.yml
ca11ab1e: Indeed, fixed :+1:

ca11ab1e:
BTW Circle CI does not run on PRs?

PR #110, opened 2022-05-27
Patch pedersen hash with cpp native implementation

THenry14:
## Usage related changes

Improve performance by patching pedersen hash calculating function from cairo-lang with c++ native code.

-

## Checklist:

- [X] Performed a self-review of the code
- [X] Rebased to the base branch
- [X] Documented the changes
- [ ] Updated the tests

## Additional details

This is a monkey patch we should eventually get rid of (when the change is upstreamed to cairo-lang itself). Using the cpp implementation of pedersen hash can yield ~30% improvement in execution time, especially for bigger, more complex contracts.

### DIFF
--- a/starknet_devnet/__init__.py
+++ b/starknet_devnet/__init__.py
@@ -1,5 +1,26 @@
 """
 Contains the server implementation and its utility classes and functions.
 """
+import sys
+from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
+from crypto_cpp_py.cpp_bindings import cpp_hash
+
 
 __version__ = "0.2.2"
+
+
+def patched_pedersen_hash(left: int, right: int) -> int:
+    """
+    Pedersen hash function written in c++
+    """
+    return cpp_hash(left, right)
+
+
+# This is a monkey-patch to improve the performance of the devnet
+# We are using c++ code for calculating the pedersen hashes
+# instead of python implementation from cairo-lang package
+setattr(
+    sys.modules["starkware.crypto.signature.fast_pedersen_hash"],
+    "pedersen_hash",
+    patched_pedersen_hash,
+)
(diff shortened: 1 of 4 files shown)


FabijanC (commented):
1) Can you show one such example where the performance is improved?
2) Can you make sure the tests are passing for you locally? You can follow .circleci/config.yml test steps to install and generate everything needed.

starknet_devnet/__init__.py
```diff
@@ -1,5 +1,23 @@
...
 """
+import sys
+from starkware.crypto.signature.fast_pedersen_hash import pedersen_hash
+from crypto_cpp_py.cpp_bindings import cpp_hash
+
 
 __version__ = "0.2.2"
+
+
+def patched_pedersen_hash(x: int, y: int) -> int:
```
FabijanC, line 12: This method doesn't comply with the project's linting standard. See yourself by running `poetry run pylint starknet_devnet test`

THenry14: Oh, thanks for pointing this out :+1: Changed in latest patchset.

THenry14:
> 1. Can you show one such example where the performance is improved?
> 
> 2. Can you make sure the tests are passing for you locally? You can follow .circleci/config.yml test steps to install and generate everything needed.


1). I have tested changes on:
- https://github.com/OilerNetwork/fossil/blob/master/contracts/starknet/FactsRegistry.cairo
Its deployment to starknet-devnet (invoking command `starknet deploy`) takes ~31 seconds. Calculating hashes takes over half of this time. After I have subsituted the default `pedersen_hash` function with one implemented in cpp, the deployment time was cut to ~20 seconds.

- I have also tested on example contract from cairo docs -> https://www.cairo-lang.org/docs/hello_starknet/intro.html#your-first-contract, but since this is very small and uncomplicated contract, there are no such gains :) 

2). I can confirm tests are passing on my linux 64bit machine; I have however some issues running on mac m1 (I have tried to run the tests without my changes - from freshly cloned repo on master branch, but they also fail, so I don’t think this is conected to the change)

FabijanC (approved):

PR #118, opened 2022-06-06
Fix events

### LINKED ISSUE #117: Events no longer show up on Receipts as of 0.2.2
When I use 0.2.1, my Starknet receipts contain the expected events.
However, when upgrading to 0.2.2, the events field is **always** an empty list on receipts.

My method and event:

```cairo

@event
func balance_increased(amount : felt):
end

# Increases the balance by the given amount.
@external
func increase_balance{
        syscall_ptr : felt*, pedersen_ptr : HashBuiltin*,
        range_check_ptr}(user : felt, amount : felt):
    let (initialized) = is_initialized.read()
    assert initialized = TRUE

    let (res) = balance.read(user=user)
    balance.write(user, res + amount)
    balance_increased.emit(amount)
    return ()
end
```

Some simple code:

```python
self.starknet_client.wait_for_tx_sync(txn_hash)
receipt = self.starknet_client.get_transaction_receipt_sync(tx_hash=txn_hash)
```

Let me know if you need more reproduction steps, but it is pretty standard contract txn i think

badurinantun:
## Usage related changes

- Fix #117  

- Test event contract invoke from the account

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/transactions.py
+++ b/starknet_devnet/transactions.py
@@ -52,7 +52,7 @@ class DevnetTransaction:
 
     def __get_events(self) -> List[Event]:
         """Returns the events"""
-        if hasattr(self.execution_info, "raw_events"):
+        if isinstance(self.execution_info, StarknetTransactionExecutionInfo):
             return self.execution_info.raw_events
 
         return self.execution_info.get_sorted_events()
--- a/starknet_devnet/transactions.py
+++ b/starknet_devnet/transactions.py
@@ -52,8 +52,10 @@ class DevnetTransaction:
 
     def __get_events(self) -> List[Event]:
         """Returns the events"""
-        contract_address = self.execution_info.call_info.contract_address
-        return [Event.create(event_content=e, emitting_contract_address=contract_address) for e in self.execution_info.call_info.events]
+        if hasattr(self.execution_info, "raw_events"):
+            return self.execution_info.raw_events
+
+        return self.execution_info.get_sorted_events()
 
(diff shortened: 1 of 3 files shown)


starknet_devnet/transactions.py
```diff
@@ -52,8 +52,10 @@ def __get_actual_fee(self) -> int:
 
     def __get_events(self) -> List[Event]:
         """Returns the events"""
-        contract_address = self.execution_info.call_info.contract_address
-        return [Event.create(event_content=e, emitting_contract_address=contract_address) for e in self.execution_info.call_info.events]
+        if isinstance(self.execution_info, StarknetTransactionExecutionInfo):
```
FabijanC, line 55: Which other class can it be other than `StarknetTransactionExecutionInfo`? And when is it so? Why doesn't that other class support using `raw_events`?

badurinantun: What other type it can be is defined by in the `execution_info` type. It can be either `TransactionExecutionInfo` for invokes or `StarknetTransactionExecutionInfo` for deploys

badurinantun: That's the only `if`. We could transform `TransactionExecutionInfo` to `StarknetTransactionExecution` but I made a priority of fixing this instead

badurinantun: [16d3eaf](https://github.com/Shard-Labs/starknet-devnet/pull/118/commits/16d3eaf20b0ef17d14e901f305b906d861ea03d0)

PR #126, opened 2022-06-13
Fix deploy through syscall

FabijanC:
## Usage related changes

- Support using the deploy syscall through accounts and in constructors.

- Expand test_declare.py

## Checklist:

- [] Performed a self-review of the code
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/starknet_wrapper.py
+++ b/starknet_devnet/starknet_wrapper.py
@@ -5,7 +5,7 @@ starkware.starknet.testing.starknet.Starknet.
 
 import dataclasses
 from copy import deepcopy
-from typing import Dict, List, Tuple
+from typing import Dict, List, Tuple, Union
 
 import cloudpickle as pickle
 from starkware.starknet.business_logic.internal_transaction import (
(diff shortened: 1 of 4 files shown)


starknet_devnet/starknet_wrapper.py
```diff
@@ -329,14 +334,19 @@ async def call(self, transaction: InvokeFunction):
 
         return { "result": adapted_result }
 
-    async def __register_new_contracts(self, execution_info: TransactionExecutionInfo):
-        for internal_call in execution_info.call_info.internal_calls:
+
+
+    async def __register_new_contracts(self, internal_calls: List[Union[FunctionInvocation, CallInfo]]):
```
FabijanC, line 339: Changed from execution_info to internal_calls to enable recursive calls (had to enable recursion to support delegate calls).

Had to use Union of types because `internal_calls` isn't of the same type for Deploy and Invoke transactions - one difference that affected Devnet is internal_call.class_hash: `int` for one type, `bytes` for the other..

Nathan-SL (approved):
LGTM!

PR #134, opened 2022-06-16
feature: added `/load` endpoint to the REST API

RedFox20:
## Usage related changes

- The `/load` endpoint allows quickly loading a previously stored Starknet chain state. This can be used for unit testing to quickly restore state between dozens of test cases
- Updated README with a simple example

- Added `load()` route into [blueprints/base.py](https://github.com/Shard-Labs/starknet-devnet/pull/134/files#diff-e36acc44798507ebeb37099b01e300126d5bd4d2c1889748ae7cd62f03d82c4a)
- Expanded `test_dump.py` to handle endpoint and non-endpoint based loading
- Added current active devnet process handling to `test_dump.py`

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/blueprints/base.py
+++ b/starknet_devnet/blueprints/base.py
@@ -32,6 +32,18 @@ def dump():
     state.dumper.dump(dump_path)
     return Response(status=200)
 
+@base.route("/load", methods=["POST"])
+def load():
+    """Loads the starknet_wrapper"""
+
+    request_dict = request.json or {}
+    load_path = request_dict.get("path")
+    if not load_path:
+        raise StarknetDevnetException(message="No path provided.", status_code=400)
+
+    state.load(load_path)
+    return Response(status=200)
+
 def validate_time_value(value: int):
     """Validates the time value"""
     if value is None:
--- a/README.md
+++ b/README.md
@@ -239,6 +239,12 @@ starknet-devnet --dump-on transaction --dump-path <PATH>
 curl -X POST http://<HOST>:<PORT>/dump -d '{ "path": <PATH> }' -H "Content-Type: application/json"
 ```
 
+- Loading state on request `/load`:
+
+```
+curl -X POST http://<HOST>:<PORT>/load -d '{ "path": <PATH> }' -H "Content-Type: application/json"
+```
+
 ### Loading
 
 To load a preserved Devnet instance, run:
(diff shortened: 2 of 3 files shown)


FabijanC (changes requested):
Hi, thanks for the PR!

One thing I would change is the testing:
Instead of having the `dump_and_load_snapshot` function with the if-else, I would use the body of that function directly in the two test cases using it (but without the if-else). Also, I would rename the functions as follows:
- `test_dumping_via_endpoint` -> `test_loading_via_cli`
- `tes_loading_via_endpoint` -> `test_dumping_and_loading_via_endpoint`

So, I'm giving you the permission to introduce code duplication (or redundancy, to make it sound better).

Also, I see you haven't marked the last check in the TODO section - can you run [all the tests locally](https://github.com/Shard-Labs/starknet-devnet/#development---test) and check the box if they pass.

RedFox20:
> One thing I would change is the testing: Instead of having the `dump_and_load_snapshot` function with the if-else, I would use the body of that function directly in the two test cases using it (but without the if-else). Also, I would rename the functions as follows:
> * `test_dumping_via_endpoint` -> `test_loading_via_cli`
> * `tes_loading_via_endpoint` -> `test_dumping_and_loading_via_endpoint`
Sure, if you prefer it like that, it is an easy change. Done.

> Also, I see you haven't marked the last check in the TODO section - can you run [all the tests locally](https://github.com/Shard-Labs/starknet-devnet/#development---test) and check the box if they pass.
I have some issues setting it up on Ubuntu, so that step is delayed. Is there a way to run these on CI ?

starknet_devnet/blueprints/base.py
```diff
@@ -32,6 +32,18 @@ def dump():
     state.dumper.dump(dump_path)
     return Response(status=200)
 
+@base.route("/load", methods=["POST"])
+def load():
+    """Loads the starknet_wrapper"""
+
+    request_dict = request.json or {}
+    load_path = request_dict.get("path") or state.dumper.dump_path
```
FabijanC, line 40: We probably don't want `state.dumper.dump_path` to be used for loading, I think the `or` and its right hand side should be removed.

FabijanC (changes requested):

FabijanC (approved):

PR #138, opened 2022-06-20
Feature/predeployed accounts endpoint

tabaktoni:
## Usage related changes
Created endpoint for retrieving devnet seeded accounts
```
/predeployed_accounts
```

- Add jsonschema as a dev dependency.

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


starknet_devnet/blueprints/base.py
```diff
@@ -88,3 +88,13 @@ async def get_balance():
         "amount": balance,
         "unit": "wei"
     })
+
+@base.route("/get_predeployed_accounts", methods=["GET"])
```
FabijanC, line 92: What do you think about keeping the verb out of the URL? Like, just having `"predeployed_accounts"`, similar to how getting account balance is done with GET /account_balance.

README.md
```diff
@@ -374,7 +374,25 @@ const account = await starknet.getAccountFromAddress(
 );
 ```
 
-The balance of an account can be checked using:
+### Predeployed accounts can be checked using:
```
FabijanC, line 377: Let's keep `##` and `###` for titles more than real sentences (so here it could be something like `### Fetch predeployed accounts`).

starknet_devnet/blueprints/base.py
tabaktoni: Agree with that, what bothers me here is the underscore instead of the dash in the API endpoint

test/support/assertions.py
```diff
@@ -0,0 +1,22 @@
+import json
+from os.path import join, dirname, abspath
+from jsonschema import validate, RefResolver
```
FabijanC, line 3: We should have `jsonschema` explicitly listed as our dev dependency.

starknet_devnet/account.py
```diff
@@ -47,6 +47,15 @@ def get_contract_class(cls):
             cls.CONTRACT_CLASS = ContractClass.load(load_nearby_contract(cls.CONTRACT_PATH))
         return cls.CONTRACT_CLASS
 
+    def to_json(self):
+        """Return json account"""
+        return {
+            "initial_balance": self.initial_balance,
```
FabijanC, line 53: Would it be better to return actual balance? It shouldn't take much longer to retrieve that, right?

test/test_account_predeployed.py
```diff
@@ -1,14 +1,48 @@
...
+def test_predeployed_accounts_endpoint():
+    """Test if the endpoint return list of predeployed accounts"""
+    response = requests.get(f"{GATEWAY_URL}/predeployed_accounts")
+    assert response.status_code == 200
+    assert_valid_schema(response.json(), "predeployed_accounts.json")
+
+
+@pytest.mark.account_predeployed
+@devnet_in_background(*ACCOUNTS_SEED_DEVNET_ARGS)
+def test_predeployed_accounts_predefined_values():
```
FabijanC, line 44: Looking at this now, I'm not sure what is the benefit of first testing the schema and then the content of the response.

FabijanC (changes requested):

starknet_devnet/account.py
tabaktoni: I just return here class properties stat is available without additional queries as part of a class in relation to the initial seed. The actual balance is out of that scope as it isn't propped but needs interaction with the chain?

FabijanC: FeeToken has a `get_balance` method that could be used. But for now we can keep it as it is, and in a next PR we could add a `balance` property (so there would be both initial_balance and balance)

test/test_account_predeployed.py
tabaktoni: This test that the preseed account method didn't change, or that the values it generated are the same as expected. There are no huge benefits but one is a test of calculate_contract_address_from_hash function.
This was a dilemma why do any test besides schema validation here.

test/support/assertions.py
tabaktoni: [e29584b1806da9f397ab618e8f7d2ddce7754f5e](https://github.com/Shard-Labs/starknet-devnet/pull/138/commits/e29584b1806da9f397ab618e8f7d2ddce7754f5e)

test/test_account_predeployed.py
FabijanC: But doesn't this second test cover everything that is tested with the first test (the schema test)?

test/support/assertions.py
tabaktoni: https://github.com/Shard-Labs/starknet-devnet/pull/138/commits/2c6f53f053bb33749b29b2948fc180c45488bf7d

test/test_account_predeployed.py
tabaktoni: https://github.com/Shard-Labs/starknet-devnet/pull/138/commits/2c6f53f053bb33749b29b2948fc180c45488bf7d

FabijanC (approved):

PR #140, opened 2022-06-24
💚 ⚡️ Fix circleci test split

tabaktoni:
## Usage related changes
None

** 💚 CircleCi Split tasks hotfix**
devnet_in_background made junitxml each object file property represented as util.py but still contained classname based on original filename without extension and with . instead of path /
You can see repercussions in (prev. test metadata):
https://app.circleci.com/pipelines/github/Shard-Labs/starknet-devnet/881/workflows/d90dbd40-4fd9-4503-90db-40674e6d4fe3/jobs/1299/parallel-runs/0/steps/0-106
To satisfy CI same tech. for classname definition is used to feed Ci split algo
Then split can decide based on provided classnames and prev. test metadata how to time-based split that classes.
After the split is done regenerate the given classnames back to filenames and feed them to pytest.
With the new working split algo total test time is now reduced to ⚡️ ~13min

**⚡️ parallelism: 2 -> 4**
Now that split works we can use parallelism to further increase test speed
with double instances, the test time is reduced by 57.14% which is close to double to ⚡️ ~8min

## Checklist:

- [x] Documented the changes
- [ ] Updated the tests
- [x] All tests are passing


.circleci/config.yml
```diff
@@ -31,18 +31,22 @@ jobs:
...
           command: |
-            TESTFILES=$(
-                circleci tests glob test/test*.py |
-                circleci tests split --split-by=timings --timings-type=classname
-            )
             mkdir -p test-results
-            poetry run pytest -s --junitxml=test-results/junit.xml -vv $TESTFILES
+            SPLITED_FILES_CLASS=$(circleci tests glob test/test*.py | sed  "s/\.py//g; s/\//./g" | circleci tests split --split-by=timings --timings-type=classname)
+            TEST_FILES=$(sed "s/\./\//g; s/ /.py /g; s/$/.py/g;" \<<< $SPLITED_FILES_CLASS)
+            poetry run pytest --verbose --junitxml=test-results/junit.xml $TEST_FILES
```
FabijanC, line 45: You've modified the `poetry run pytest` command:
- Can you restore the `-s` flag?
- Is there any difference between `-vv` and `--verbose`?

.circleci/config.yml
```diff
@@ -31,18 +31,22 @@ jobs:
...
           name: Run pytest tests
           parallel: true
           command: |
-            TESTFILES=$(
-                circleci tests glob test/test*.py |
-                circleci tests split --split-by=timings --timings-type=classname
-            )
             mkdir -p test-results
-            poetry run pytest -s --junitxml=test-results/junit.xml -vv $TESTFILES
+            SPLITED_FILES_CLASS=$(circleci tests glob test/test*.py | sed  "s/\.py//g; s/\//./g" | circleci tests split --split-by=timings --timings-type=classname)
```
FabijanC, line 43: - Why does this `sed` have two replace scripts, and the one in the next line has three? Is it because SPLITED_FILES_CLASS is a one-line string?
- Why are you using SPLITED_FILES_CLASS as an intermediary variable? Can't the output of `circleci test split` be piped into `sed`? Then maybe the next line could have just two replace scripts.
- Just in case, can you change `s/\.py//g` to `s/\.py$//g` so that we're sure that it only alters the extension.

.circleci/config.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     docker:
       - image: cimg/python:3.7.12-node
     resource_class: medium
-    parallelism: 2
+    parallelism: 4
```
FabijanC, line 19: If the total execution time for `parallelism: 4` is `8*4=32` minutes, aren't we paying more credits than with `parallelism: 2` (`13*2=26` minutes)

FabijanC (changes requested):

.circleci/config.yml
tabaktoni: - Yes
- It can, but with an intermediate variable, it is more readable what's going on. It still need to have three, one to restore / one to retore .py to all words 'except last one because it has no space' and one to replace to last one.
- It will not work because only the last word will receive an extension. This is one line.

.circleci/config.yml
tabaktoni: Yes, we play a little bit more for double the speed, not the double pay, as for 18 min there are ~3 in setup. There is no free lunch, but this is close as it gets. And with more tests coming we had no worry for some time.

.circleci/config.yml
```diff
@@ -31,18 +31,22 @@ jobs:
       - run:
           name: Compile test contracts
           command: ./scripts/compile_contracts.sh
+      - run:
+          name: Check out previous test metadata
+          command: | 
+            cat "${CIRCLE_INTERNAL_TASK_DATA}/circle-test-results/results.json" | jq .
```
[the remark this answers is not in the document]
tabaktoni: I just copy-pasted this from CI docs.

.circleci/config.yml
tabaktoni: I will restore -s -vv, as -vv is double verbose aka. max verbose

.circleci/config.yml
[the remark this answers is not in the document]
tabaktoni: https://github.com/Shard-Labs/starknet-devnet/pull/140/commits/a36cb40adaafa764d07966725410d5b00b98d8cd

FabijanC (approved):

PR #165, opened 2022-07-06
⚡️ Implement xdist for parallel testing

tabaktoni:
Introduce parallel testing for local development or SSH machine split testing
``` poetry run pytest -v -n auto --dist loadscope test/```

You can specify n number of workers that will execute the test.
```-n auto```

My machine is a Dual-Core Intel Core i5  so ```-n auto``` will set a number of workers based on a number of cores to 2.
However, based on my test I get the best performance when set ```-n 4``` so each core deal with two workers.

- ⚡️ local test from 30+ min -> 19 min (Dual-Core Intel Core i5)
- package added pytest-xdist

## Checklist:

- [x] No linter errors
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/test/settings.py
+++ b/test/settings.py
@@ -9,7 +9,6 @@ def bind_free_port(host):
     port = str(sock.getsockname()[1])
     return port, f"http://{host}:{port}"
 
-
 HOST = "127.0.0.1"
 PORT, APP_URL = bind_free_port(HOST)
 
--- a/poetry.lock
+++ b/poetry.lock
@@ -1213,7 +1213,7 @@ testing = ["pytest (>=6)", "pytest-checkdocs (>=2.4)", "pytest-flake8", "pytest-
 [metadata]
 lock-version = "1.1"
 python-versions = ">=3.7.2,<3.10"
-content-hash = "e18959e79b52ded2c4b8b8f6583019a135e5397a0754d7495375446553fb2b4a"
+content-hash = "24ee88e402dddfa9522d2d379f693494e1ca064d302f811c75fe9bb1fc16944e"
 
 [metadata.files]
 aiohttp = [

### CODE REVIEW

File: test/settings.py [hunk 1]
```diff
@@ -1,11 +1,28 @@
...
-L1_HOST = "localhost"
-L1_PORT = "8545"
-
-APP_URL = f"http://{HOST}:{PORT}"
-GATEWAY_URL = APP_URL
-FEEDER_GATEWAY_URL = APP_URL
-L1_URL = f"http://{L1_HOST}:{L1_PORT}/"
+import socket
+
+# pylint: disable=invalid-name,too-few-public-methods
```
FabijanC, line 5: Why is invalid-name check suppressed? Perhaps it can really be a better name, like NetworkSettings.

pyproject.toml
```diff
@@ -25,6 +25,7 @@ pylint = "~2.12.2"
 web3 = "~5.28.0"
 psutil = "~5.9.1"
 jsonschema = "~3.2.0"
+pytest-xdist = "^2.5.0"
```
FabijanC, line 28: Can you add it with a tilde instead of a caret?

README.md
```diff
@@ -511,6 +511,8 @@ When running tests locally, do it from the project root:
 
 poetry run pytest test/
 
+poetry run pytest -n auto --dist loadscope test/  # parallel testing using n CPU workers
```
FabijanC, line 514: The docs say:
> Tests are grouped by module for test functions and by class for test methods. 

Just to be clear, currently all our tests are in the form of test functions, right?

test/settings.py
```diff
@@ -1,11 +1,28 @@
...
+    APP_URL = ""
+
+    L1_HOST = "localhost"
+    L1_PORT = ""
+    L1_URL = ""
+
+    def __init__(self):
+        '''Setup app endpoints for testing'''
+        sock = socket.socket()
+        sock.bind(('', 0))
```
FabijanC, line 19: Does it get automatically unbounded when the process dies?

test/settings.py
```diff
@@ -1,11 +1,28 @@
...
-
-APP_URL = f"http://{HOST}:{PORT}"
-GATEWAY_URL = APP_URL
-FEEDER_GATEWAY_URL = APP_URL
-L1_URL = f"http://{L1_HOST}:{L1_PORT}/"
+import socket
+
+# pylint: disable=invalid-name,too-few-public-methods
+class Settings:
+    '''settings'''
```
FabijanC, line 7: Can we use the same docstring quoting across the whole project?

test/settings.py
```diff
@@ -1,11 +1,28 @@
...
-L1_PORT = "8545"
-
-APP_URL = f"http://{HOST}:{PORT}"
-GATEWAY_URL = APP_URL
-FEEDER_GATEWAY_URL = APP_URL
-L1_URL = f"http://{L1_HOST}:{L1_PORT}/"
+import socket
+
+# pylint: disable=invalid-name,too-few-public-methods
+class Settings:
```
FabijanC, line 6: Any advantage why it's a class that gets instantiated and not a class with a class method? Also, why even convert this into a class and not keep it as root-level constants?

test/settings.py
```diff
@@ -1,11 +1,28 @@
...
+    PORT = ""
+    APP_URL = ""
+
+    L1_HOST = "localhost"
+    L1_PORT = ""
+    L1_URL = ""
+
+    def __init__(self):
+        '''Setup app endpoints for testing'''
+        sock = socket.socket()
```
FabijanC, line 18: Can we extract lines 18-20 and 23-25 to a function called `get_free_port` or `bind_free_port`?

test/settings.py
FabijanC, line 19: We should use double quotes uniformly, right?

FabijanC (changes requested):

README.md
tabaktoni: Yes, so at the end it split it by files same as on CI

test/settings.py
tabaktoni: Yes sorry, this is my habit to use single-quoted originally from PHP because PHP interpreter needs to interpret double-quoted, later I used it also in js single doesn't need to escape html node prop.

test/settings.py
tabaktoni: Do you mean like the static method? Because then it cant change self and I need like 'singleton'.
Because as root level dev can import var ex. PORT in test and in runtime port will be changed to designated port but the imported PORT is not by reference so it will stay unchanged. So to avoid that originally I created get_port() and set_port() functions that can then be used to fix reference problems as they will be exec in runtime and get new value but still you could make mistake using the imported PORT for ex.
Then I try to create something like JS getter and setter for Vars but can't find python way to make it as in JS so the class was the final solution to make it clean and unable for dev to make mistake.

test/settings.py
tabaktoni: Agree,  My 10-year habit of always using single quotes

test/settings.py
FabijanC: Three approaches here:
1) You can use a `@staticmethod` which access class properties directly by class name, not through `self`
2) You can use a `@classmethod` which has access to `cls` instead of `self`.
3) As suggested here: https://github.com/Shard-Labs/starknet-devnet/pull/165#discussion_r915755873, the port can be assigned directly if used with the helper function, so no need for accessing properties after their definition.

FabijanC (approved):

PR #166, opened 2022-07-07
/api endpoint to get existing endpoints

tabaktoni:
## Usage related changes

3rd party test for existing endpoints

## Checklist:

- [x] No linter errors
- [ ] Documented the changes
- [ ] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -111,17 +111,18 @@ def handle(error: StarkException):
     """Handles the error and responds in JSON. """
     return {"message": error.message, "status_code": error.status_code}, error.status_code
 
-@app.route('/api', methods = ['GET'])
+@app.route("/api", methods = ["GET"])
 def api():
-    """Print available endpoints."""
+    """Return available endpoints."""
     routes = {}
     # pylint: disable=protected-access
     for url in app.url_map._rules:
-        routes[url.rule] = {}
-        routes[url.rule]["functionName"] = url.endpoint
-        routes[url.rule]["methods"] = list(url.methods)
-        routes[url.rule]["doc"] = app.view_functions[url.endpoint].__doc__
-    routes.pop("/static/<path:filename>")
+        routes[url.rule] = {
+            "functionName": url.endpoint,
+            "methods": list(url.methods),
+            "doc": app.view_functions[url.endpoint].__doc__
+        }
+    #routes.pop("/static/<path:filename>")
     return jsonify(routes)
 
 if __name__ == "__main__":
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -5,7 +5,7 @@ A server exposing Starknet functionalities as API endpoints.
 import sys
 from pickle import UnpicklingError
 
-from flask import Flask
+from flask import Flask, jsonify
 from flask_cors import CORS
 import meinheld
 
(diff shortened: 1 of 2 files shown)


starknet_devnet/server.py
```diff
@@ -111,5 +111,18 @@ def handle(error: StarkException):
     """Handles the error and responds in JSON. """
     return {"message": error.message, "status_code": error.status_code}, error.status_code
 
+@app.route('/api', methods = ['GET'])
+def api():
+    """Print available endpoints."""
+    routes = {}
+    # pylint: disable=protected-access
+    for url in app.url_map._rules:
+        routes[url.rule] = {}
```
FabijanC, line 120: Can't it be constructed right away, without further assignments, like:
```python
routes[url.rule] = {
  "functionName": ...,
  ...
}
```

starknet_devnet/server.py
```diff
@@ -111,5 +111,18 @@ def handle(error: StarkException):
     """Handles the error and responds in JSON. """
     return {"message": error.message, "status_code": error.status_code}, error.status_code
 
+@app.route('/api', methods = ['GET'])
+def api():
+    """Print available endpoints."""
```
FabijanC, line 116: Does it print or return the endpoints?

FabijanC (changes requested):
You can add a test for this in the test_endpoints.py file. Doesn't have to be anything strict, but would be good to have something that at least runs the api function.

The PR description says you're adding a "3rd party test for existing endpoints". In what sense is it 3rd party and in what sense is it a test?

starknet_devnet/server.py
```diff
@@ -111,5 +111,18 @@ def handle(error: StarkException):
...
+    routes = {}
+    # pylint: disable=protected-access
+    for url in app.url_map._rules:
```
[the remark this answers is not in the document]
tabaktoni: WSGI https://werkzeug.palletsprojects.com/en/2.1.x/routing/#werkzeug.routing.Map and used by flask https://wsgi.readthedocs.io/en/latest/frameworks.html

tabaktoni:
> The PR description says you're adding a "3rd party test for existing endpoints". In what sense is it 3rd party and in what sense is it a test?

As a developer using starknet-devnet in his own project, he can create a test that will check for all existing endpoints that starknet-devnet implements. If starknet-devnet changes some endpoints, his tests will fail and he will know he needs to fix his app for the new version.

starknet_devnet/server.py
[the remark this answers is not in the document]
tabaktoni: Oh about the response, I am not aware of some sort of existing standard

FabijanC, line 119: Instead of disabling protected access linter rule, we could use `app.url_map.iter_rules()`

starknet_devnet/server.py
```diff
@@ -111,5 +111,18 @@ def handle(error: StarkException):
...
+        routes[url.rule] = {}
+        routes[url.rule]["functionName"] = url.endpoint
+        routes[url.rule]["methods"] = list(url.methods)
+        routes[url.rule]["doc"] = app.view_functions[url.endpoint].__doc__
```
FabijanC, line 123: I observed the output that we get for some endpoints. E.g. `/feeder_gateway/get_code` has this value under the `"doc"` property:
```
"\n    Returns the ABI and bytecode of the contract whose contractAddress is provided.\n    "
```
We could `.strip()` the `__doc__`

FabijanC (approved):
Looks great! Except for the single quotes in the test, but I can do a quick push.

PR #186, opened 2022-07-19
Support JSON-RPC write API

### LINKED ISSUE #172: Support JSON-RPC write API

drknzz:
## Usage related changes

Added support for [JSON-RPC write API](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json) 

None

## Checklist:

- [x] No linter errors
- [ ] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

(closes #172)


starknet_devnet/blueprints/rpc.py
```diff
@@ -330,6 +338,76 @@ async def get_events():
...
+    )
+
+
+async def add_declare_transaction(contract_class: RpcContractClass, version: str) -> dict:
+    """
+    Submit a new class declaration transaction
+    """
+    try:
+        decompressed_program = decompress_program({"contract_class": contract_class}, False)["contract_class"]
+        contract = ContractClass.loads(json.dumps(decompressed_program, sort_keys=True))
```
FabijanC, line 366: Shouldn't it be called `contract_class` rather than `contract`?

starknet_devnet/blueprints/rpc.py
```diff
@@ -2,19 +2,27 @@
...
+from __future__ import annotations
+import dataclasses
 import json
-from typing import Callable, Union, List, Tuple, Optional, Any
 
+from typing import Callable, Union, List, Tuple, Optional, Any
 from typing_extensions import TypedDict
-
 from flask import Blueprint, request
+import marshmallow
```
FabijanC, line 14: This should be made a dependency, we shouldn't be relying on a third party to install it.

starknet_devnet/blueprints/rpc.py
```diff
@@ -2,19 +2,27 @@
...
-from __future__ import annotations
+# pylint: disable=too-many-lines
 
+from __future__ import annotations
+import dataclasses
 import json
-from typing import Callable, Union, List, Tuple, Optional, Any
 
+from typing import Callable, Union, List, Tuple, Optional, Any
 from typing_extensions import TypedDict
```
FabijanC, line 12: Even though this is an old contribution, this should be made a dependency, we shouldn't be relying on a third party to install it.

starknet_devnet/blueprints/rpc.py
FabijanC, line 366: Also, why `sort_keys=True`? And isn't it simpler to do `ContractClass.load(decompressed_program)` - so `.load` instead of `.loads`?

starknet_devnet/blueprints/rpc.py
```diff
@@ -905,9 +1006,11 @@ def parse_body(body: dict) -> Tuple[Callable, Union[List, dict], int]:
         "getClassHashAt": get_class_hash_at,
         "getClassAt": get_class_at,
         "estimateFee": estimate_fee,
+        "addInvokeTransaction": add_invoke_transaction,
+        "addDeclareTransaction": add_declare_transaction,
+        "addDeployTransaction": add_deploy_transaction,
     }
-
-    method_name = body["method"].lstrip("starknet_")
+    method_name = body["method"].replace("starknet_", "")
```
FabijanC, line 1013: Why did this become necessary?

starknet_devnet/blueprints/rpc.py
```diff
@@ -2,19 +2,27 @@
 RPC routes
 rpc version: 0.15.0
 """
-from __future__ import annotations
+# pylint: disable=too-many-lines
```
FabijanC, line 5: This is a reminder that we need to refactor :smiley:

starknet_devnet/blueprints/rpc.py
```diff
@@ -330,6 +338,76 @@ async def get_events():
...
+
+
+async def add_declare_transaction(contract_class: RpcContractClass, version: str) -> dict:
+    """
+    Submit a new class declaration transaction
+    """
+    try:
+        decompressed_program = decompress_program({"contract_class": contract_class}, False)["contract_class"]
+        contract = ContractClass.loads(json.dumps(decompressed_program, sort_keys=True))
+        contract = dataclasses.replace(contract, abi=[])
```
FabijanC, line 367: Why is this necessary?

starknet_devnet/blueprints/rpc.py
```diff
@@ -330,6 +338,76 @@ async def get_events():
...
+        class_hash=rpc_felt(class_hash),
+    )
+
+async def add_deploy_transaction(contract_address_salt: str, constructor_calldata: List[str], contract_definition: RpcContractClass) -> dict:
+    """
+    Submit a new deploy contract transaction
+    """
+    try:
+        decompressed_program = decompress_program({"contract_definition": contract_definition}, False)["contract_definition"]
+        contract = ContractClass.loads(json.dumps(decompressed_program, sort_keys=True))
```
FabijanC, line 392: Can you call this either `contract_class` or `contract_definition`?

test/rpc/test_rpc_transactions.py
```diff
@@ -320,3 +323,125 @@ def test_get_deploy_transaction_receipt(deploy_info):
...
+        params={
+            "function_invocation": function_invocation,
+            "signature": invoke_content["signature"],
+            "max_fee": hex(0),
+            "version": hex(constants.TRANSACTION_VERSION),
+        }
+    )
+    receipt = resp["result"]
+
+    assert set(receipt.keys()) == set(["transaction_hash"])
```
FabijanC, line 349: Here and in other places we are doing just key assertions, no values are being asserted. This can be done either how it's done in other rpc tests, or you could use `assert_valid_schema` as it's done in `test_account_predeployed.py`

FabijanC (changes requested):

starknet_devnet/blueprints/rpc.py
drknzz: We cannot get abi from compressed program string, and without it [this line](https://github.com/starkware-libs/cairo-lang/blob/13cef109cd811474de114925ee61fd5ac84a25eb/src/starkware/starknet/testing/contract_utils.py#L169) throws an assertion error.

starknet_devnet/blueprints/rpc.py
drknzz: lstrip() works different than expected here.
Here are the [docs](https://python-reference.readthedocs.io/en/latest/docs/str/lstrip.html).

Especially this example shows why the change was necessary:
```
>>> "ABBA".lstrip("AB") # both AB and BA are stripped
''
```

So basically when `starknet_addDeployTransaction` was in `body["method"]`, the result was `ddDeployTransaction`.

test/rpc/test_rpc_transactions.py
drknzz: I'm having a hard time getting expected values for these assertions.

The class hashes here are different than when computed using gateway.
Same goes for transaction hash, probably since it's missing abi in contract_definition.

Could you point me to any source for the expected results?

FabijanC: yeah, we don't have to assert the precise value, but at least the structure of the response (i.e. that the value is a hex string).

pyproject.toml
```diff
@@ -19,6 +19,8 @@ meinheld = "~1.0.2"
 Werkzeug = "~2.0.3"
 cloudpickle = "~2.1.0"
 crypto-cpp-py = "~1.0.4"
+marshmallow = "^3.17.0"
```
FabijanC, line 22: Please redo the installation of these packages by doing the tilde (~) installation. It has to be done with `poetry add` rather than manually, otherwise the content-hash in poetry.lock doesn't get calculated properly which prevents later docker image building. Speaking of the content-hash, I see it's reported by GitHub as conflicting with master, you should probably rebase/merge.

FabijanC (changes requested):

pyproject.toml
drknzz: Should I keep these versions but with a tilde instead of caret symbol? Wouldn't the content-hash be changed then?
I did install these two packages using `poetry add`
It's the same situation as [here](https://stackoverflow.com/questions/62378386/how-can-i-add-a-dependency-in-poetry-using-tilde-instead-of-caret).

FabijanC: Doing `poetry add` but with a tilde should work (and it would also correctly update the content-hash): so `poetry add marshmallow~3.17.0`

starknet_devnet/blueprints/rpc.py
drknzz: It does indeed replace ContractClass's default `None` abi value with `[]`.
And the assertion in `get_abi` only requires for it not to be `None`.

What bugs me is that ContractClass's `abi` in its [definition](https://github.com/starkware-libs/cairo-lang/blob/13cef109cd811474de114925ee61fd5ac84a25eb/src/starkware/starknet/services/api/contract_class.py#L47) is actually `Optional`.

poetry.lock
```diff
@@ -1329,7 +1329,9 @@ cachetools = [
     {file = "cachetools-5.2.0-py3-none-any.whl", hash = "sha256:f9f17d2aec496a9aa6b76f53e3b614c965223c061982d434d160f930c698a9db"},
     {file = "cachetools-5.2.0.tar.gz", hash = "sha256:6a94c6402995a99c3970cc7e4884bb60b4a8639938157eeed436098bf9831757"},
 ]
-cairo-lang = []
+cairo-lang = [
+    {file = "cairo-lang-0.9.1.zip", hash = "sha256:a34cc1a019330d8eabf70869fd3d92a530af26de70844a42037db4679ac0e1e0"},
```
FabijanC, line 1333: Why did this change happen?

PR #197, opened 2022-07-27
Fix rpc declare and deploy parameters

cptartur:
## Usage related changes

- Add type cast to rpc declare transaction `version` parameter. [RPC spec specifies ](https://github.com/starkware-libs/starknet-specs/blob/master/api/starknet_write_api.json#L75)that the `version` should be be a number represented as hex value, however current rpc implementation doesn't include required casts.
- Add type cast to rpc deploy calldata and salt parameters.

- Removed incomplete typeddict call from `add_invoke_transaction` test.

## Checklist:

- [x] No linter errors
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/test/rpc/test_rpc_transactions.py
+++ b/test/rpc/test_rpc_transactions.py
@@ -365,7 +365,7 @@ def test_add_declare_transaction_on_incorrect_contract(declare_content):
         "starknet_addDeclareTransaction",
         params={
             "contract_class": rpc_contract,
-            "version": constants.TRANSACTION_VERSION,
+            "version": hex(constants.TRANSACTION_VERSION),
         }
     )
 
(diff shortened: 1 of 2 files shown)


test/rpc/test_rpc_transactions.py
```diff
@@ -7,7 +7,7 @@
 from typing import List
 
 from starkware.starknet.definitions import constants
-from starknet_devnet.blueprints.rpc import RpcContractClass, RpcInvokeTransaction
```
FabijanC, line 10: After removing this import, is `RpcInvokeTransaction` used anywhere at all?

cptartur: It is used in the rpc.py blueprint, for example in the `rpc_invoke_transaction()` function.

FabijanC: Ah right, it's used for type specification.

FabijanC (approved):

PR #201, opened 2022-08-01
Fix/get transaction status

tabaktoni:
## Usage related changes

Fix get_transaction_status block_hash to be hex-string

GET ```/feeder_gateway/get_transaction_status```
mainet and testnet response:

```
{
    "block_hash": "0x76b...",
    "tx_status": "ACCEPTED_ON_L1"
}
```

devnet response:

```
{
    "block_hash": 10178...,
    "tx_status": "ACCEPTED_ON_L2"
}
```

## Checklist:

- [x] No linter errors
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


starknet_devnet/blueprints/feeder_gateway.py
```diff
@@ -162,6 +162,8 @@ def get_transaction_status():
 
     transaction_hash = request.args.get("transactionHash")
     transaction_status = state.starknet_wrapper.transactions.get_transaction_status(transaction_hash)
+    if "block_hash" in transaction_status:
```
FabijanC, line 165: Shouldn't this be handled in `DevnetTransactions.get_transaction_status`? There we already have one `if` which checks if `"block_hash"` exists.

test/test_endpoints.py
```diff
@@ -248,3 +255,21 @@ def test_error_response_class_by_hash():
...
+    })
+    assert response.status_code == 200
+    tx_hash = response.json().get("tx_hash")
+
+    json_response = get_transaction_status(tx_hash)
+    assert_valid_schema(json_response, "get_transaction_status.json")
+
+    invalid_tx_hash = "0x443a8b3ec1f9e0c64"
+    json_response = get_transaction_status(invalid_tx_hash)
+    assert_valid_schema(json_response, "get_transaction_status.json")
```
FabijanC, line 275: Are we happy with just testing the response of a valid and invalid request the same way?

FabijanC (changes requested):

test/test_endpoints.py
tabaktoni: What can be a more elegant way?

Testnet return for non-existing address 
```
{
    "tx_status": "NOT_RECEIVED"
}
```
This test checks that we return it also.
Single JSON Schema should provide all cases with a common schema interface for one endpoint.

FabijanC: Asserting valid schema is good; perhaps an extra line could be added after the schema assertion that checks the exact value of `"tx_status"`.

starknet_devnet/blueprints/feeder_gateway.py
tabaktoni: Whatever I touch it opens a new construction site 👷 , so I wanted to avoid it, but you are right, will fix it the proper way.

tabaktoni: b58288c9f2b2959f516250c2ae767ee8d989706a

test/test_endpoints.py
tabaktoni: 4c7527bfd06be4f2abdd193ff9de7219b7b69b08

PR #207, opened 2022-08-03
Replace meinheld with gunicorn

Solpatium:
## Usage related changes

1. Access log has the same format and is printed to stderr as it was before.
2. No logs were changed.

1. Replaced [meinheld](https://github.com/mopemope/meinheld) with [gunicorn](https://gunicorn.org/).

Reasons:
- meinheld seems to be abandoned. Last changes are 2 years old and https://meinheld.org/ is some blog now (?).
- devnet [terminated in starknet.py's CI](https://github.com/software-mansion/starknet.py/runs/7632895291?check_suite_focus=true#step:8:155) with a message `*** stack smashing detected ***: terminated`. After replacing it with gunicorn it [doesn't happen anymore](https://github.com/software-mansion/starknet.py/runs/7647401088?check_suite_focus=true). Other users found [other buffer-related issues](https://github.com/mopemope/meinheld/issues/120).

3. Renamed debug script to match naming convention of others and name in README.
4. Used poetry in debug script to use poetry environment with local dependenies.

## Checklist:

- [x] No linter errors
- [ ] Documented the changes - **NA**
- [ ] Linked the issues which this PR resolves - **NA**
- [ ] Updated the tests - **NA**
- [x] All tests are passing


starknet_devnet/server.py
```diff
@@ -85,6 +84,26 @@ def set_gas_price(args):
...
+    def __init__(self, application, args):
+        self.args = args
+        self.application = application
+        super().__init__()
+
+    def load_config(self):
+        self.cfg.set("bind", f"{self.args.host}:{self.args.port}")
+        self.cfg.set("workers", 1)
+        # "-" means stdout
+        self.cfg.set("accesslog", "-")
```
FabijanC, line 101: So far this has been piped to stderr.

starknet_devnet/server.py
```diff
@@ -102,9 +121,8 @@ def main():
     set_gas_price(args)
 
     try:
-        meinheld.listen((args.host, args.port))
         print(f" * Listening on http://{args.host}:{args.port}/ (Press CTRL+C to quit)")
```
FabijanC, line 124: Instead of removing this line, could we remove all the gunicorn [INFO] logs? They appear on startup and on SIGINT.

starknet_devnet/server.py
```diff
@@ -85,6 +84,26 @@ def set_gas_price(args):
...
+    """Our Gunicorn application."""
+
+    def __init__(self, application, args):
+        self.args = args
+        self.application = application
+        super().__init__()
+
+    def load_config(self):
+        self.cfg.set("bind", f"{self.args.host}:{self.args.port}")
+        self.cfg.set("workers", 1)
```
FabijanC, line 99: How does it handle simultaneous requests?

FabijanC (changes requested):
Surely meinheld was ready to be replaced with something documented and maintained; thanks for the PR and for fixing the other issues. I left a few notes.

starknet_devnet/server.py
Solpatium: It queues the requests just like meinheld did and servers them.
I run 7 simultaneous requests for get_balance just to make sure.

**meinheld** version:
image

**gunicorn** version:
image

Code I used:
```
x = () => fetch("http://127.0.0.1:5050/account_balance?address=ACCOUNT"
Promise.all([x(), x(), x(), x(), x(), x(), x()])
```

We can't have more than one worker as the state between them is not shared.

FabijanC: Great, thanks for the explanation and visualization!

Solpatium: Calls to contracts could potentially be run simultaneously, but probably not with flask, it starts an event loop per request. Do users complain about it?

FabijanC: I remember something like that being mentioned some half a year ago.

FabijanC:
Sorry, there was an issue with the lock file, can you resolve the conflicts again?

starknet_devnet/server.py
Solpatium: It is should be possible to optimise calls by executing them asynchronously on a CPU bound executor or by forking the process, but this is a discussion for [other](https://github.com/Shard-Labs/starknet-devnet/issues/208) probably.

starknet_devnet/server.py
Solpatium: It should behave the same now :)

FabijanC: Great work!

FabijanC (approved):

PR #219, opened 2022-08-10
Isolate devnet config, move parsing, dump synchronously

### LINKED ISSUE #218: Dump/load no longer works on 0.2.9
**Describe the bug**
It seems that the state is reset.

**Not reproducible on alpha-goerli**

- [x] This issue is only present on Devnet and cannot be reproduced on alpha-goerli (check the box if true).

**To Reproduce**

1. Start up the devnet using `-dump-path file -dump-on-exit` 
2. Exit 
3. Start up the devnet using `-dump-path file -dump-on-exit --load-path file` 
4. Deploy a wallet using `starknet deploy_account --account test`
5. Check the byte code : `starknet get_code --contract_address [the address]`
6. Exit and restart devnet as in 3.7. 
7. Check again as in 5 -> it's empty.

**Expected behavior**
In 0.2.8 and earlier, the byte code is returned, indicating the state is lost.

**Devnet version**

Work on 0.2.6, 0.2.7, 0.2.8
Broken on 0.2.9

**Machine specifications**

- OS: Mac 12.3.1
- CPU: M1 arm
- Memory: 64G

FabijanC:
## Usage related changes

- Fix dump/load and restart functionalities (fix #218)
- Wait for dump to finish (potentially a slowdown, but should eliminate race-related bugs)

- Predeploy 1 account by default in tests (instead of 0 - this caused headache while trying to figure out why tests were passing but real-world was failing)
- Make rpc tests more independent of each other:
  - Set rpc fixtures module to function instead of scope
  - Restart state after each test
- Refactor code:
  - Expand functionality of `DevnetConfig` and move it from `starknet_wrapper.py` to `devnet_config.py`
  - Move `parse_args` from `util.py` to `devnet_config.py`
  - Moved account generation from `server.py` to `__init__` of `Accounts`.

## Checklist:

- [x] No linter errors
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -2,7 +2,6 @@
 A server exposing Starknet functionalities as API endpoints.
 """
 
-from pickle import UnpicklingError
 import sys
 import asyncio
 
@@ -11,6 +10,8 @@ from flask_cors import CORS
 from gunicorn.app.base import BaseApplication
 from starkware.starkware_utils.error_handling import StarkException
 
+from .util import StarknetDevnetException
+
 from .starknet_wrapper import StarknetWrapper
 
 from .blueprints.base import base
@@ -18,7 +19,6 @@ from .blueprints.gateway import gateway
 from .blueprints.feeder_gateway import feeder_gateway
 from .blueprints.postman import postman
 from .blueprints.rpc import rpc
-from .util import check_valid_dump_path
 from .state import state
 from .devnet_config import DevnetConfig, DumpOn, parse_args
 
@@ -37,28 +37,9 @@ app.register_blueprint(feeder_gateway)
 app.register_blueprint(postman)
 app.register_blueprint(rpc)
 
-def set_dump_options(args):
-    """Assign dumping options from args to state."""
-    if args.dump_path:
-        try:
-            check_valid_dump_path(args.dump_path)
-        except ValueError as error:
-            sys.exit(str(error))
-
-    state.dumper.dump_path = args.dump_path
-    state.dumper.dump_on = args.dump_on
-
-def load_dumped(args):
-    """Load a previously dumped state if specified."""
-    if args.load_path:
-        try:
-            state.load(args.load_path)
-        except (FileNotFoundError, UnpicklingError):
-            sys.exit(f"Error: Cannot load from {args.load_path}. Make sure the file exists and contains a Devnet dump.")
-
 # We don't need init method here.
 # pylint: disable=W0223
-class Devnet(BaseApplication):
+class GunicornServer(BaseApplication):
     """Our Gunicorn application."""
 
     def __init__(self, application, args):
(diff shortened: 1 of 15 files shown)


starknet_devnet/state.py
```diff
@@ -2,30 +2,30 @@
...
-        self.dumper = Dumper(self.starknet_wrapper)
+        self.set_starknet_wrapper(StarknetWrapper(DevnetConfig()))
 
-    def __set_starknet_wrapper(self, starknet_wrapper: StarknetWrapper):
+    def set_starknet_wrapper(self, starknet_wrapper: StarknetWrapper):
         """Sets starknet wrapper and creates new instance of dumper"""
         self.starknet_wrapper = starknet_wrapper
         self.dumper = Dumper(starknet_wrapper)
 
-    async def reset(self, config: DevnetConfig = None):
```
FabijanC, line 21: This parameter was never used.

starknet_devnet/state.py
```diff
@@ -2,30 +2,30 @@
...
     """
     Stores starknet wrapper and dumper
     """
     def __init__(self):
-        self.starknet_wrapper = StarknetWrapper(config=DevnetConfig())
-        self.dumper = Dumper(self.starknet_wrapper)
+        self.set_starknet_wrapper(StarknetWrapper(DevnetConfig()))
 
-    def __set_starknet_wrapper(self, starknet_wrapper: StarknetWrapper):
+    def set_starknet_wrapper(self, starknet_wrapper: StarknetWrapper):
```
FabijanC, line 16: Made public so it can be used from `main` in `server.py`.

starknet_devnet/starknet_wrapper.py
```diff
@@ -71,6 +65,11 @@ def __init__(self, config: DevnetConfig):
         self.fee_token = FeeToken(self)
         self.accounts = Accounts(self)
 
+        if config.start_time is not None:
```
FabijanC, line 68: Moved here from server.py

starknet_devnet/server.py
```diff
@@ -123,22 +92,17 @@ def load_config(self):
...
 
-    args = parse_args()
-
     # Uncomment this once fork support is added
     # origin = Origin(args.fork) if args.fork else NullOrigin()
     # starknet_wrapper.origin = origin
 
+    args = parse_args(sys.argv[1:])
+    state.set_starknet_wrapper(StarknetWrapper(DevnetConfig(args)))
     load_dumped(args)
```
tabaktoni, line 104: what do you think about?
state.dumper.load_dumped(args)
state.dumper.set_dump_options(args)

It implicates move dumper functions from server to dump.py

starknet_devnet/server.py
```diff
@@ -123,22 +92,17 @@ def load_config(self):
...
     """Runs the server."""
 
-    args = parse_args()
-
     # Uncomment this once fork support is added
     # origin = Origin(args.fork) if args.fork else NullOrigin()
     # starknet_wrapper.origin = origin
 
+    args = parse_args(sys.argv[1:])
+    state.set_starknet_wrapper(StarknetWrapper(DevnetConfig(args)))
```
tabaktoni, line 103: It is called two times when you start an app on state init and after that here again.

tabaktoni (approved):
Like this direction in separation of concerns

starknet_devnet/server.py
FabijanC: Doesn't affect performance, tests using app.test_client don't work without it.

starknet_devnet/server.py
FabijanC: https://github.com/Shard-Labs/starknet-devnet/pull/219/commits/95372005781725c71157cd5ac74ce9e644e9542b

PR #227, opened 2022-08-23
Test/rpc/constructor calldata

### LINKED ISSUE #198: Test RPC `add_deploy_transaction` with `constructor_calldata != []`
Contract currently used in RPC deploy transaction test currently does not accept any constructor_calldata, which means we are not testing if we are correctly parsing passed calldata. This should be improved by either changing contract used in these tests or adding a separate tests using different contract with nonempty constructor parameters.

internnos:
## Usage related changes

no change for user

closes #198 

changed `test/deploy_rpc.json`
- added constructor calldata
- changed ABI to match ABI output from `test/contracts/cairo/contract.cairo`
- changed `entry_points_by_type` to match compilation output from `test/contracts/cairo/contract.cairo`

changed `starknet_devnet/blueprints/rpc/call.py`
- added handling invalid input for calldata (before executing starknet `call`)

### generating `test/deploy_rpc.json`
`abi` is generated from
```
nile compile test/contracts/cairo/contract.cairo
```

`program` is generated from
```
from starkware.starknet.compiler.compile import compile_starknet_files
from starkware.starknet.services.api.gateway.transaction_utils import compress_program, decompress_program

contract_class = compile_starknet_files(['test/contracts/cairo/contract.cairo'])
program = compress_program(contract_class.program.Schema().dump(contract_class.program))
```

## Checklist:

- [x] No linter errors
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


test/rpc/test_rpc_class.py
```diff
@@ -20,12 +20,27 @@ def test_get_class(class_hash):
     contract_class = resp["result"]
 
     assert contract_class["entry_points_by_type"] == {
-        "CONSTRUCTOR": [],
-        "EXTERNAL": [
-            {"offset": "0x3a", "selector": "0x0362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320"},
-            {"offset": "0x5b", "selector": "0x039e11d48192e4333233c7eb19d10ad67c362bb28580c604d67884c85da39695"}
-        ],
-        "L1_HANDLER": []
+        "CONSTRUCTOR": [
```
ivpavici, line 23: Can this be extracted somehow instead of having it copied in 2 places?
https://github.com/Shard-Labs/starknet-devnet/pull/227/files#diff-ffa0bbaa73d5171327e5293e41e360ddfe123c2c8a0bbf47dd27c48b948ec569R81

internnos: oh yeah i think it'd neater that way

on it!

starknet_devnet/blueprints/rpc/call.py
```diff
@@ -23,6 +23,9 @@ async def call(request: FunctionCall, block_id: BlockId) -> List[Felt]:
         raise RpcError(code=20, message="Contract not found")
 
     try:
+        for calldata in request["calldata"]:
+            if not isinstance(calldata, str):
```
FabijanC, line 27: - Why not check if it's a felt and not just a string (as per the specifications: https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L1442)
The logic could be extracted to a function (you could apply `hex` and catch `ValueError`).

- I don't like the fact that `calldata` is used as the name of an item of `request["calldata"]`.

- Is this mentioned in the PR description?

- Looking at `test_call_raises_on_invalid_calldata`, looks like this was already covered, right? So I guess the motivation for introducing this extra check is just to detect bad input earlier?

test/rpc/test_rpc_class.py
```diff
@@ -62,13 +78,6 @@ def test_get_class_at(deploy_info):
...
-        "EXTERNAL": [
-            {"offset": "0x3a", "selector": "0x0362398bec32bc0ebb411203221a35a0301193a96f317ebe5e40be9f60d15320"},
-            {"offset": "0x5b", "selector": "0x039e11d48192e4333233c7eb19d10ad67c362bb28580c604d67884c85da39695"}
-        ],
-        "L1_HANDLER": []
-    }
+    assert contract_class["entry_points_by_type"] == EXPECTED_CONTRACT_CLASS
     assert isinstance(contract_class["program"], str)
-    decompress_program({"contract_class": contract_class}, False)
+    decompress_program({"contract_class": contract_class}, False)
```
FabijanC, line 83: The linter should warn about the missing newline

test/rpc/test_rpc_call.py
```diff
@@ -100,6 +100,33 @@ def test_call_raises_on_invalid_calldata(deploy_info):
     }
 
 
+@pytest.mark.usefixtures("run_devnet_in_background")
+def test_call_raises_on_invalid_calldata_datatype_calldata_should_be_empty(deploy_info):
```
FabijanC, line 104: The name of this test is not clear to me and it is quite long. You can also rely on the function docstring.

test/rpc/test_rpc_call.py
```diff
@@ -88,7 +88,7 @@ def test_call_raises_on_invalid_calldata(deploy_info):
             "request": {
                 "contract_address": pad_zero(contract_address),
                 "entry_point_selector": hex(get_selector_from_name("get_balance")),
-                "calldata": ["a", "b", "123"],
```
FabijanC, line 91: Is this necessary?

FabijanC (changes requested):
I unchecked the linter point in your PR checklist. Please run the linter (`./scripts/lint.sh`), correct the reported errors, and recheck it in the list.

starknet_devnet/blueprints/rpc/call.py
internnos: > Why not check if it's a felt and not just a string (as per the specifications: https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json#L1442)
The logic could be extracted to a function (you could apply hex and catch ValueError).

i think this is good idea

> I don't like the fact that calldata is used as the name of an item of request["calldata"].
i can change this

> Looking at test_call_raises_on_invalid_calldata, looks like this was already covered, right? So I guess the motivation for introducing this extra check is just to detect bad input earlier?

yes this is for this point

> Is this mentioned in the PR description?
nope but this throws error after i changed the deploy contract

on it!

test/rpc/test_rpc_call.py
internnos: i think yes, because the checking needs to be as rigorous as possible

1. check if the datatype does not comply
2. check if the datatype is correct, but doesn't comply to the number of expected parameters supplied

what do you think about this?

@FabijanC

test/rpc/test_rpc_call.py
internnos: gotcha!

starknet_devnet/blueprints/rpc/call.py
FabijanC: > actually no. so do i need to remove this?

Just update the PR description

test/rpc/test_rpc_class.py
```diff
@@ -80,4 +80,5 @@ def test_get_class_at(deploy_info):
 
     assert contract_class["entry_points_by_type"] == EXPECTED_CONTRACT_CLASS
     assert isinstance(contract_class["program"], str)
-    decompress_program({"contract_class": contract_class}, False)
\ No newline at end of file
+    decompress_program({"contract_class": contract_class}, False)
+
```
FabijanC, line 84: I still see something reported here

starknet_devnet/blueprints/rpc/call.py
internnos: alrite!

test/rpc/test_rpc_call.py
FabijanC: Sure we need to test with invalid input, but I don't understand this two-item checklist that you provided - is that something you claim this particular test is doing? Or is it related to other tests as well (and how)? Perhaps pytest.mark.parametrize could be of use (already used in some rpc tests)

internnos: oh yeah i think i should try to use `pytest.mark.parametrize`

thank you!

test/rpc/test_rpc_class.py
FabijanC: This commit introduced the required change: https://github.com/Shard-Labs/starknet-devnet/pull/227/commits/ebc7f2f99a15e8a6ee8d28db5fbe5e5a9335bd23
This commit undid the change: https://github.com/Shard-Labs/starknet-devnet/pull/227/commits/6be3079bb5cd839444feaf89d17751a79a63a998

starknet_devnet/blueprints/rpc/call.py
internnos: done!

FabijanC:
You didn't properly rebase

internnos:
> You didn't properly rebase

done!

FabijanC:
Something is still wrong, for an example, if you take a look at the changed files section (https://github.com/Shard-Labs/starknet-devnet/pull/227/files), you'll see changes not done by you.

internnos:
> https://github.com/Shard-Labs/starknet-devnet/pull/227/files

i have resolved it

thank you!

internnos:
> Something is still wrong, for an example, if you take a look at the changed files section (https://github.com/Shard-Labs/starknet-devnet/pull/227/files), you'll see changes not done by you.

i have resolved it

thank you!

starknet_devnet/blueprints/rpc/call.py
```diff
@@ -22,13 +22,22 @@ async def call(request: FunctionCall, block_id: BlockId) -> List[Felt]:
     if not state.starknet_wrapper.contracts.is_deployed(int(request["contract_address"], 16)):
         raise RpcError(code=20, message="Contract not found")
 
+    # exception handling using [rpc_felt(int(calldata_value, 1)) for calldata_value in request["calldata"]]
+    # does not work properly, so we use the following workaround
+    for calldata_value in request["calldata"]:
+        try:
+            rpc_felt(int(calldata_value, 16))
+        except (ValueError, TypeError):
+            raise RpcError(code=22, message="Invalid calldata") from None
```
FabijanC, line 31: Instead of raising from `None`, you should do `except (ValueError, TypeError) as error`, and raise from `error`. Is `except`ing both `ValueError` and `TypeError` necessary?

README.md
```diff
@@ -121,6 +121,7 @@ The latest stable version is also available as:
 
 - `shardlabs/starknet-devnet:latest`
 
+
```
FabijanC, line 124: Where did this come from?

test/rpc/test_rpc_call.py
```diff
@@ -88,15 +90,15 @@ def test_call_raises_on_invalid_calldata(deploy_info):
...
+                "calldata": calldata,
             },
             "block_id": "latest"
         }
     )
-
+
```
FabijanC, line 98: Doesn't the linter report this extra whitespace?

test/rpc/conftest.py
```diff
@@ -54,7 +54,6 @@ def fixture_invoke_info() -> dict:
     Make an invoke transaction on devnet and return invoke info dict
     """
     invoke_tx = json.loads(INVOKE_CONTENT)
-    invoke_tx["calldata"] = ["0"]
```
FabijanC, line 57: Why aren't other tests affected by this?

FabijanC (changes requested):
Can you make sure you linted the code and that the tests are passing?

test/rpc/conftest.py
internnos: the invoke test was not that much in the first place i think,

since all tests are currently passed

FabijanC:
The commit history looks wrong (indicative of rebasing/merging problems). Ideally I shouldn't be able to see my profile picture in the list of commits of this PR.

starknet_devnet/blueprints/rpc/call.py
internnos: great idea! will try it

yeah i think this is necessary since `TypeError: int() can't convert non-string with explicit base` might occur for 
`["wtf"]`

FabijanC:
> yeah there were some problems
> 
> if all are resolved, i think i will create a fresh PR from main
> 
> what do you think?

You can create a new PR, especially since #237 was merged (which introduced more conflicts)

test/rpc/conftest.py
FabijanC: Still don't understand, why was the setting of the calldata needed before this PR, but not anymore?

PR #233, opened 2022-08-25
Refactor/1 empty block creation

### LINKED ISSUE #200: Refactor empty block creation
Introduced with #189, empty block creation is done in a method separate from the already existent `generate_block`. Empty block generation could be framed as a case of `generate_block`.

internnos:
## Usage related changes

CLOSES #200 

No change for user

Refactor empty block creation

## Checklist:

- [x] No linter errors
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


test/deploy_rpc.json
```diff
@@ -44,4 +44,5 @@
     },
     "contract_address_salt": "0x2",
     "type": "DEPLOY"
-}
+
+}
```
ivpavici, line 48: add new line after the bracket :)

internnos: done!

starknet_devnet/starknet_wrapper.py
```diff
@@ -160,7 +160,7 @@ async def __store_transaction(
         Stores the provided data as a deploy transaction in `self.transactions`.
         Generates a new block
         """
-
+
```
ivpavici, line 163: tab/spaces here

starknet_devnet/blocks.py
```diff
@@ -151,51 +158,4 @@ async def generate(
...
-                block_hash=block_hash,
-                old_root=state_update.old_root,
-                new_root=state_update.new_root,
-                state_diff=state_update.state_diff,
-            )
-
-        self.__state_updates[block_number] = state_update
-
-        return block
+        return block
```
ivpavici, line 161: missing newline

starknet_devnet/starknet_wrapper.py
```diff
@@ -421,4 +423,4 @@ def set_block_time(self, time_s: int):
 
     def set_gas_price(self, gas_price: int):
         """Sets gas price to `gas_price`."""
-        self.block_info_generator.set_gas_price(gas_price)
+        self.block_info_generator.set_gas_price(gas_price)
```
ivpavici, line 426: missing newline

starknet_devnet/blocks.py
```diff
@@ -120,15 +126,16 @@ async def generate(
                 tx_signatures=[signature],
                 event_hashes=[],
                 sequencer_address=state.general_config.sequencer_address
-            )
-
-        transaction_receipts = (transaction.get_execution(),)
+                )
```
ivpavici, line 129: here indentation for the whole calculate_block_hash?

starknet_devnet/blocks.py
```diff
@@ -99,18 +99,24 @@ async def generate(
...
             parent_block_hash = last_block.block_hash
 
-        if self.lite:
+        if is_empty_block:
+            transaction_receipts = ()
+            transactions = []
             block_hash = block_number
         else:
-            block_hash = await calculate_block_hash(
+
```
ivpavici, line 114: new line here?

ivpavici:
Please double check about linter errors!

internnos:
aah gotcha! sorry i forgot to run the linter

internnos:
can i use black to lint automatically? @ivpavici

starknet_devnet/blocks.py
internnos: yes that is correct!

starknet_devnet/blocks.py
```diff
@@ -120,15 +124,15 @@ async def generate(
...
                 event_hashes=[],
                 sequencer_address=state.general_config.sequencer_address
-            )
-
-        transaction_receipts = (transaction.get_execution(),)
+                )
 
+            transaction_receipts = (transaction.get_execution(),)
+            transactions=[transaction.internal_tx]
         block = StarknetBlock.create(
```
FabijanC, line 131: How about an empty line here?

starknet_devnet/blocks.py
FabijanC: The indentation might be syntactically correct, but the readability could be improved by adding an extra tab to the `calculate_block_hash` arguments. However, this might be irrelevant if the comment on the line 113 is resolved.

starknet_devnet/blocks.py
```diff
@@ -99,18 +99,22 @@ async def generate(
...
             parent_block_hash = last_block.block_hash
 
-        if self.lite:
+        if is_empty_block:
+            transaction_receipts = ()
+            transactions = []
             block_hash = block_number
         else:
-            block_hash = await calculate_block_hash(
+            if self.lite:
```
FabijanC, line 113: Couldn't nested ifs somehow be avoided? By having something like `if is_empty_block or self.lite`?

FabijanC (changes requested):

starknet_devnet/blocks.py
internnos: oh sure!

starknet_devnet/blocks.py
internnos: how about this?

starknet_devnet/blocks.py
```diff
@@ -99,85 +99,40 @@ async def generate(
...
             parent_block_hash = last_block.block_hash
 
-        if self.lite:
-            block_hash = block_number
+        if is_empty_block:
+            transaction_receipts = ()
+            transactions = []
         else:
+            signature = transaction.get_signature()
             block_hash = await calculate_block_hash(
```
FabijanC, line 113: The arguments to this function are still not indented (tabbed).

starknet_devnet/blocks.py
```diff
@@ -99,85 +99,40 @@ async def generate(
...
-            parent_block_hash = 0
-        else:
-            last_block = self.__get_last_block()
-            parent_block_hash = last_block.block_hash
-
-        #Fake block number
-        block_hash = block_number
+            transaction_receipts = (transaction.get_execution(),)
+            transactions=[transaction.internal_tx]
+        if self.lite or is_empty_block:
```
FabijanC, line 126: Would be good to have empty lines between adjacent if-blocks.

starknet_devnet/blocks.py
```diff
@@ -99,85 +99,40 @@ async def generate(
...
         else:
             last_block = self.__get_last_block()
             parent_block_hash = last_block.block_hash
 
-        if self.lite:
-            block_hash = block_number
+        if is_empty_block:
+            transaction_receipts = ()
+            transactions = []
         else:
```
FabijanC, line 111: This `else` will be run even in lite mode (basically every time that `is_empty_block` is not `True`), which is not the previous behavior.

FabijanC (changes requested):
Do you run tests before each push? `./scripts/test.sh`

I'm asking because with the last push, changes were introduced which don't just refactor but also change the behavior of the refactored code. If you ran the tests and the tests didn't report an error, that means the tests are not good.

starknet_devnet/blocks.py
internnos: aah i see, should have used the the more explicit elif then

what do you think?

FabijanC: not sure if just replacing `else` with `elif CONDITION` would be enough

internnos: i have pushed it, kindly check if you have the time @FabijanC

thank you!

starknet_devnet/blocks.py
internnos: > 

i ran the tests and they were not reporting anything except `too many arguments`

Screen Shot 2022-08-26 at 9 03 06 am
g



> Do you run tests before each push? `./scripts/test.sh`
> 
> I'm asking because with the last push, changes were introduced which don't just refactor but also change the behavior of the refactored code. If you ran the tests and the tests didn't report an error, that means the tests are not good.

yes, i always ran the tests. unless its green i won't push

starknet_devnet/blocks.py
FabijanC: The point of lite mode is to avoid unnecessary calculation, like calculating block hash. This wasn't handled properly in the code which I commented on and the calculation was done even in lite mode. So functionally, the behavior was the same, but not performance-wise (it was slower). But it looks better after the changes, I'll quickly do a formal review.

internnos: yes i agree, performance wise, in the commit before this, it was slower because of the reason you have pointed!

but now i think the latest commit resolves this issue.

thanks for your review!

starknet_devnet/blocks.py
```diff
@@ -90,7 +90,7 @@ def get_state_update(self, block_hash=None, block_number=None) -> BlockStateUpda
 
     async def generate(
         self, transaction: DevnetTransaction, state: StarknetState,
-        state_root: bytes, state_update = None
+        state_root: bytes, state_update = None, is_empty_block = False
```
FabijanC, line 93: If the linter is reporting extra arguments here, it means it exited with a non-zero status code and the CI/CD is going to fail. One possible fix is to put a pylint suppression just for this function, but maybe even better would be to exclude the `state_root` argument and extract it from `state` inside this function, if possible.

FabijanC (changes requested):
The logic looks good now, just some style/lint points remain.

starknet_devnet/starknet_wrapper.py
```diff
@@ -285,6 +284,8 @@ async def deploy(self, deploy_transaction: Deploy) -> Tuple[int, int]:
             transaction_hash=tx_hash,
         )
 
+
```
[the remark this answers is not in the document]
internnos: gotcha!

starknet_devnet/blocks.py
internnos: alrite! on it!

internnos: i've inspect the `state` but it doesn't seem to contain `state_root`

i think i'll suppress the linter specifically for this function

FabijanC: Look at the usage of the generate function and what's passed into it, should be something like `state.state.shared_state.contract_states`

internnos: oh yeah thats correct! 

thank you!

internnos: there's even a getter function for that, apology for missing that function

done!

FabijanC: Great, I see the getter can be removed now that it's not used. Will approve and merge.

FabijanC (approved):

Issue #253, opened 2022-09-07
Research if block hash can be deterministic for test purposes.


mikiw:
Research if block hash can be deterministic for test purposes. If yes modify test_general_workflow with parameterized hash block.

mikiw:
https://docs.starknet.io/docs/Blocks/header

mikiw:
Yes, block hashes can be deterministic but 2 conditions need to be met:
- Timestamps need to be the same, so a new mechanism for that needs to be implemented to be machine independent in tests
-  Everything that requires salt (transactions, address in contract deploy, state) requires salt to be deterministic not random.

mikiw:
The motivation of that task was to avoid this if in one test:
```
    # check if in lite mode expected block hash is 0x1
    if expected_block_hash == EXPECTED_SALTY_DEPLOY_BLOCK_HASH_LITE_MODE:
        assert_equal(expected_block_hash, get_block(parse=True)["block_hash"])
```
in
test_general_workflow.py

FabijanC:
@mikiw Thanks for the research, closing then.

PR #255, opened 2022-09-07
Unification of test_general_workflow test

### LINKED ISSUE #250: Unification of test_general_workflow test
test_general_workflow need's to be parameterized

mikiw:
## Usage related changes

Unification of test_general_workflow test, closes #250

- from now on in lite mode ```tx_hash = internal_tx.hash_value``` because it is calculated anyway, this might change in future

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


test/conftest.py
```diff
@@ -9,10 +9,18 @@
...
+def fixture_expected_tx_hash(request):
     """
-    Fixture to return values of expected hash
+    Fixture to return values of expected tx hash
+    """
+    return request.param
+
+
+@pytest.fixture(name="expected_block_hash")
+def fixture_expected_block_hash(request):
```
FabijanC, line 21: Is this used anywhere? If these are used in just one file, they can be defined there instead of in conftest

test/test_general_workflow.py
```diff
@@ -65,6 +71,10 @@ def test_general_workflow(expected_hash):
     # check block and receipt after deployment
     assert_negative_block_input()
 
+    # check if in lite mode expected block hash is 0x1
```
FabijanC, line 74: do you mean this as a TODO for later?

test/test_deploy.py
```diff
@@ -87,8 +87,12 @@ async def test_deploy_lite():
         contract_class=deploy_transaction.contract_definition,
     )
 
+    # Currently in lite mode hashes are actually calculated
+    assert (
+        tx_hash
+        == 2009361157170034697813403089970408499498259480420986282652588934221678003889
+    )
     assert contract_address == expected_contract_address
-    assert tx_hash == 0
```
FabijanC, line 96: As agreed, this test can be parameterized.

FabijanC (changes requested):

test/test_general_workflow.py
mikiw: No, it will be as it is. Block hash is 0x1. I just created a new task for block problem #253.

If it will be possible to fix that I'll fix that. If not I'll just leave it as it is because it will be only possible to check it in lite mode.

test/conftest.py
mikiw: I was also planning to use this in test_block_number_incremented and test_deploy_lite but in the future.

You know version 2.0 (but for real not kidding).

test/test_general_workflow.py
FabijanC: So if #253 is resolved successfully, we can remove the `if`. Resolving this comment.

FabijanC (approved):

PR #281, opened 2022-09-21
Update CI; fix seed logging

### LINKED ISSUE #211: When poetry 1.2.0 is released, add poetry lock check to CI/CD
https://github.com/python-poetry/poetry/issues/1406

Motivation comes from wanting to be able to tell right on CI/CD start if the lock file is correct. Currently we find that out in the middle of the image building process when it fails (ungracefully and after all the tests have potentially successfully passed). This should be addressed with parallelization of image building in #230

### LINKED ISSUE #280: Remove the --rm option when testing docker container
If the Devnet inside the container fails, the container is removed immediately and the logs are inaccessible.

FabijanC:
## Usage related changes

- Looks like #264 was not done properly, now it should be fixed.

- Since poetry 1.2 is being installed in CI/CD, we can now use the `poetry lock --check` functionality which it introduces.
  - Close #211 
- Improve image building
  - Remove `--rm` so that docker logs can be accessed in case of early exit  - close #280 
  - Refactor - better use of variables, not relying on hardcoded port

## Checklist:

- [x] All tests are passing

### DIFF
--- a/scripts/install_dev_tools.sh
+++ b/scripts/install_dev_tools.sh
@@ -13,4 +13,5 @@ echo "poetry: $(poetry --version)"
 
 # install dependencies
 poetry install
+poetry lock --check
 npm ci
--- a/starknet_devnet/accounts.py
+++ b/starknet_devnet/accounts.py
@@ -19,7 +19,11 @@ class Accounts:
         self.starknet_wrapper = starknet_wrapper
         self.__n_accounts = starknet_wrapper.config.accounts
         self.__initial_balance = starknet_wrapper.config.initial_balance
+
         self.__seed = starknet_wrapper.config.seed
+        if self.__seed is None:
+            self.__seed = random.getrandbits(32)
+
         self.list = []
 
         self.__generate()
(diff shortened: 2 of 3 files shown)


starknet_devnet/accounts.py
```diff
@@ -42,11 +46,9 @@ def add(self, account):
     def __generate(self):
         """Generates accounts without deploying them"""
         random_generator = random.Random()
-        self.__initial_balance = self.__initial_balance
+        random_generator.seed(self.__seed)
 
-        random_generator.seed(
-            self.__seed if self.__seed is not None else random_generator.getrandbits(32)
-        )
+        self.__initial_balance = self.__initial_balance
```
mikiw, line 51: Looks suspicious to me.

scripts/install_dev_tools.sh
```diff
@@ -13,4 +13,5 @@ echo "poetry: $(poetry --version)"
 
 # install dependencies
 poetry install
+poetry lock --check
```
mikiw, line 16: Please explain why we need that. I'm poetry noob.

starknet_devnet/accounts.py
FabijanC: https://github.com/Shard-Labs/starknet-devnet/pull/281/commits/7685c2c09a14763cfd2cdfd17637806040c122bd

Issue #282, opened 2022-09-23
issue with positional parameter in JSON-RPC with starknet_addInvokeTransaction


gregoryguillou:
When I run the following RPC call with positional parameter and `starknet_addInvokeTransaction`, I got an error saying `int() can't convert non-string with explicit base`. This is not happening with other verbs (at least those I have tested). I cannot really workaround that issue because the go ethereum JSON-RPC client does not support positional parameters. 

This is the request that is failing:

```shell
curl -H 'Content-Type: application/json' \
-d '{
    "id": 6,
    "jsonrpc": "2.0",
    "method": "starknet_addInvokeTransaction",
    "params": [
        {
            "calldata": [
                "0x1",
                "0x35a55a64238b776664d7723de1f6b50350116a1ab1ca1fe154320a0eba53d3a",
                "0x7a44dde9fea32737a5cf3f9683b3235138654aa2d189f6fe44af37a61dc60d",
                "0x0",
                "0x0",
                "0x0",
                "0x0"
            ],
            "contract_address": "0x6bb9425718d801fd06f144abb82eced725f0e81db61d2f9f4c9a26ece46a829",
            "entry_point_selector": "0x15d40a3d6ca2ac30f4031e42be28da9b056fef9bb7357ac5e85627ee876e5ad"
        },
        [
            "0x43d30855910c7923b4ebd55c5b51edb2b8a220a0ddd6c3b71f69e7f55f7b838",
            "0x40f65b1dc85eea253de3d7c29a7ae6a77f4ab18e30ccc0255a8fe97e5f2c679"
        ],
        "0x47e82c3a41000",
        "0x0"
    ]
}' http://localhost:5050/rpc
```

And the error:

```text
{"error":{"code":22,"message":"int() can't convert non-string with explicit base"},"id":6,"jsonrpc":"2.0"}
```

If I do the exact same request with named parameters it just works. This is the conversion:

```shell
curl -H 'Content-Type: application/json' \
-d '{
    "id": 6,
    "jsonrpc": "2.0",
    "method": "starknet_addInvokeTransaction",
    "params": {
        "function_invocation": {
            "calldata": [
                "0x1",
                "0x35a55a64238b776664d7723de1f6b50350116a1ab1ca1fe154320a0eba53d3a",
                "0x7a44dde9fea32737a5cf3f9683b3235138654aa2d189f6fe44af37a61dc60d",
                "0x0",
                "0x0",
                "0x0",
                "0x0"
[...]

PR #283, opened 2022-09-23
Fix positional args for addInvokeTransaction

gregoryguillou:
## Usage related changes
This is definitely a bug in devnet that is not showing up in goerli (pathfinder). We should be able to call the RPC api with positional argument and those arguments should come in the right order, i.e. signature comes 2nd in the case of addInvokeTransaction. Check the v0.1.0 spec for details

I guess there is a good reason why you have set the signature back with None with both default and Optional. However, it prevents any positional arguments from working and you would want to comply to it. I assume Pathfinder requires you pass a an empty list if they do not want to provide a signature. The test cases are documented in #282

N/A

## Checklist:

- [ ] Documented the changes <- What you expect I do?
- [x] Updated the tests
- [ ] All tests are passing <- One test on L1/L2 is failing on my laptop. I assume it is unrelated and I am missing some setup/node version.

### DIFF
--- a/starknet_devnet/blueprints/rpc/transactions.py
+++ b/starknet_devnet/blueprints/rpc/transactions.py
@@ -106,10 +106,10 @@ async def pending_transactions() -> List[RpcTransaction]:
 
 async def add_invoke_transaction(
     function_invocation: FunctionCall,
+    signature: List[Felt],
     max_fee: NumAsHex,
     version: NumAsHex,
     nonce: NumAsHex = None,
-    signature: Optional[List[Felt]] = None,
 ) -> dict:
     """
     Submit a new transaction to be added to the chain
--- a/starknet_devnet/blueprints/rpc/transactions.py
+++ b/starknet_devnet/blueprints/rpc/transactions.py
@@ -3,7 +3,7 @@ RPC transaction endpoints
 """
 
 import dataclasses
-from typing import List, Optional
+from typing import List
 
 from marshmallow.exceptions import MarshmallowError
 from starkware.starknet.services.api.contract_class import ContractClass
--- a/test/rpc/test_rpc_transactions.py
+++ b/test/rpc/test_rpc_transactions.py
@@ -313,6 +313,32 @@ def test_add_invoke_transaction(invoke_content):
     assert set(receipt.keys()) == {"transaction_hash"}
     assert receipt["transaction_hash"][:3] == "0x0"
 
+@pytest.mark.usefixtures("run_devnet_in_background")
+def test_add_invoke_transaction_positional_args(invoke_content):
+    """
+    Add invoke transaction
+    """
+    resp = rpc_call(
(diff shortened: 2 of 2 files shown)


starknet_devnet/blueprints/rpc/transactions.py
```diff
@@ -106,10 +106,10 @@ async def pending_transactions() -> List[RpcTransaction]:
 
 async def add_invoke_transaction(
     function_invocation: FunctionCall,
+    signature: List[Felt],
```
FabijanC, line 109: `Optional` means the value can be `None` and is used when the default value is `None`. Parameters with a default value are defined last. The specification states that the signature is not a required parameter:
https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_write_api.json#L23

I think the problem is also that JSON-RPC API v0.1.0 is not compatible with the new transaction version (1) and that this could be addressed with adaptation to API v0.2.0.
@drknzz what do you say, would it make sense to accept the solution in this PR as a temporary fix?

gregoryguillou: Thank you,

The [addInvokeTransaction spec for v0.1 in RPC](https://github.com/starkware-libs/starknet-specs/blob/30e5bafcda60c31b5fb4021b4f5ddcfc18d2ff7d/api/starknet_write_api.json#L9) says the signature should come 2nd. The fact that it comes 5 after the Nonce is an issue for the Go SDK for the following reasons:
- Go only supports positional parameters, not named parameters
- Pathfinder have the parameter in the 2nd place already so I would need to implement different version of the protocol depending on the API provider.

I sure understand that in order to have the parameter optional, you need to push it to the end. Minimum I would prefer you put it in front of the Nonce that is not even supposed to exist in v0.1. I understand you are trying to make things better for everyone but, to me, leaving the signature in 2nd and making is mandatory (and possibly empty) is a better choice than pushing it in 5th. What do you think?

drknzz: There's some point in what @gregoryguillou is saying, so I think it could be accepted rn.

Mark that in RPC v0.2.0 the interface [changes](https://github.com/starkware-libs/starknet-specs/compare/v0.1.0...v0.2.0?diff=unified#diff-b10ba31e7dae56dbcf6f0b7d1c9d9e708b212a037c02b72741f276a04d9c14f4R11) though:

```
"name": "starknet_addInvokeTransaction",
"summary": "Submit a new transaction to be added to the chain",
"params": [
    {
        "name": "invoke_transaction",
        "description": "The information needed to invoke the function (or account, for version 1 transactions)",
        "required": true,
        "schema": {
            "$ref": "#/components/schemas/BROADCASTED_INVOKE_TXN"
        }
    }
]
```

FabijanC (approved):

PR #295, opened 2022-10-03
Starknet get events rpc

mikiw:
## Usage related changes

- getEvents RPC added

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -3,13 +3,64 @@
...
+        (is_empty(keys)) or (is_not_empty(keys) and bool(set(event.keys) & set(keys)))
+    )
+
+
+def get_events_from_block(block, address, keys):
+    """
+    Return filtered events.
+    """
+    events = []
+    for event in block.transaction_receipts[0].events:
```
tabaktoni, line 57: Note: implementation with one transaction per block

test/rpc/test_rpc_misc.py
```diff
@@ -121,3 +150,113 @@ def test_call_with_invalid_params(params):
...
+                "keys": [
+                    "0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9",
+                    "0x3db3da4221c078e78bd987e54e1cc24570d89a7002cefa33e548d6c72c73f9d",
+                ],
+            },
+            2,
+        ),
+    ],
+    indirect=True,
+)
```
tabaktoni, line 249: To be honest I don't like this parameterized input as it makes the test file overall unclean.
It is an elegant solution but if it can be extracted into the single events_parameters config file and imported and pass like single PARAMETERS, it would be much cleaner for the test file ?

test/rpc/test_rpc_misc.py
```diff
@@ -121,3 +150,113 @@ def test_call_with_invalid_params(params):
...
+    """
+    deploy_info = deploy(EVENTS_CONTRACT_PATH)
+    for i in range(0, 2):
+        invoke(
+            calls=[(deploy_info["address"], "increase_balance", [i])],
+            account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
+            private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+        )
+    resp = rpc_call("starknet_getEvents", params=input_data)
+    assert len(resp["result"]["events"]) == expected_data
```
tabaktoni, line 262: This ensures that the number of events returned is the number of events expected.
As we are not testing events but filtering I agree that we don't need to check Event structure, but also that the same filtering can return the right number of events but wrong ones (miss order, broken mapping, new starknet version break)?
Could we check that we get exactly the events we expect to get for each case?

tabaktoni (changes requested):

starknet_devnet/blueprints/rpc/misc.py
mikiw: I thought that currently we support only one transaction in block

test/rpc/test_rpc_misc.py
mikiw: ok, will do

test/rpc/test_rpc_misc.py
mikiw: I was thinking about that also, one time I have a false positive like this when I was implementing chunking.

I'll try to figure out something better.

ivpavici:
Docs should be updated also:

```
Methods currently not supported:

starknet_protocolVersion - will be removed in a future version of the specification
**starknet_getEvents**
```

test/rpc/test_rpc_misc.py
mikiw: Changed

test/rpc/test_rpc_misc.py
mikiw: Changed for checking if events data are equal

starknet_devnet/blueprints/rpc/misc.py
mikiw: Changed for all events in all transaction_receipts

mikiw:
@ivpavici I have updated json-rpc-api.md

tabaktoni (approved):

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +78,48 @@ async def syncing() -> Union[dict, bool]:
...
 
 
-# pylint: disable=redefined-builtin
-async def get_events(filter: dict) -> dict:
+# pylint: disable=too-many-arguments
+async def get_events(
+    from_block: BlockId,
+    to_block: BlockId,
+    address: Address = "",
+    keys: List[Address] = Empty,
```
FabijanC, line 86: Why use `queue.Empty` here? Isn't `None` more appropriate here?

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +78,48 @@ async def syncing() -> Union[dict, bool]:
...
+
+    In our implementation continuation_token is just a number.
+
+    In state.starknet_wrapper.get_state().events there is no relation between blocks.
+    This is why we need to iterate block by block, take all events,
+    and chunk it later which is not an optimal solution.
     """
-    raise NotImplementedError()
+    events = []
+    keys = [] if keys == Empty else [int(k, 0) for k in keys]
```
FabijanC, line 100: When comparing with singletons, the `is` operator should be used in favor of `==`: https://stackoverflow.com/a/14247419

test/rpc/test_rpc_misc.py
```diff
@@ -121,3 +150,25 @@ def test_call_with_invalid_params(params):
...
+    "run_devnet_in_background, input_data, expected_data",
+    GET_EVENTS_TEST_DATA,
+    indirect=True,
+)
+def test_get_events(input_data, expected_data):
+    """
+    Test RPC get_events.
+    """
+    deploy_info = deploy(EVENTS_CONTRACT_PATH)
+    for i in range(0, 2):
```
FabijanC, line 166: The more usual way is to do `for i in range(2)`

test/rpc/test_rpc_misc.py
```diff
@@ -121,3 +150,25 @@ def test_call_with_invalid_params(params):
...
+        invoke(
+            calls=[(deploy_info["address"], "increase_balance", [i])],
+            account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
+            private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+        )
+    resp = rpc_call("starknet_getEvents", params=input_data)
```
FabijanC, line 172: What if returned data contains more elements than expected? Since iterating over expected data, that will not be registered, right? Perhaps adding a length assertion would make sense.

test/rpc/test_data/get_events.py
```diff
@@ -0,0 +1,119 @@
...
+    "keys": [
+        "0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9",
+        "0x3db3da4221c078e78bd987e54e1cc24570d89a7002cefa33e548d6c72c73f9d",
+    ],
+}
+
+EVENT_DATA_0 = "[0, 0]"
+EVENT_DATA_1 = "[0, 1]"
+EVENT_FEE = (
+    "[1483697464188569092488551792700745002099692778312526338015109648680562628122, "
```
FabijanC, line 69: This looks very ugly, the formatter probably doesn't touch it because it's a single multiline string. Perhaps you could do call `str(...)` on an actual object defined with `[...]`.

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +78,48 @@ async def syncing() -> Union[dict, bool]:
...
+    to_block = (
+        state.starknet_wrapper.blocks.get_number_of_blocks()
+        if to_block == "latest"
+        else to_block
+    )
+
+    for block_number in range(int(from_block), int(to_block)):
```
FabijanC, line 107: This stops iteration before reaching `to_block`, is that the intended behavior? Nothing is specified [here](https://github.com/starkware-libs/starknet-specs/blob/v0.1.0/api/starknet_api_openrpc.json), right?

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +78,48 @@ async def syncing() -> Union[dict, bool]:
     return False
 
 
-# pylint: disable=redefined-builtin
-async def get_events(filter: dict) -> dict:
+# pylint: disable=too-many-arguments
+async def get_events(
+    from_block: BlockId,
+    to_block: BlockId,
+    address: Address = "",
```
FabijanC, line 85: If `""` bears some special meaning, it should be extracted, and the same constant should be used in all places in the code.

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -3,13 +3,64 @@
...
-
-from starknet_devnet.blueprints.rpc.structures.payloads import Felt, Address
 from starknet_devnet.state import state
 
 
+def is_empty(value):
+    """
+    Return true if empty.
+    """
+    return not value and isinstance(value, collections.Iterable)
```
FabijanC, line 26: I don't see the need for such an emptiness check. The most pythonic way is to use the object itself. Like, instead of doing `if is_empty(obj):` just doing `if obj`. What is the risk in not doing it that way?

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -3,13 +3,64 @@
...
+
+
+def is_not_empty(value):
+    """
+    Return true if not empty.
+    """
+    return not is_empty(value)
+
+
+def filter_address(address, event):
```
FabijanC, line 36: These `filter_` functions have odd names. I would expect them to actually filter something. Whereas by looking into their code, it seems they are just indicators if something should be filtered or not.

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -3,13 +3,64 @@
...
+    """
+    return bool(address == "" or event.from_address == int(address, 0))
+
+
+def filter_keys(keys, event):
+    """
+    Filter by keys.
+    """
+    return bool(
+        (is_empty(keys)) or (is_not_empty(keys) and bool(set(event.keys) & set(keys)))
```
FabijanC, line 48: Unnecessary complication with boolean operators; `x V (!x ^ y)` can be simplified to `x V y`:
https://www.wolframalpha.com/input?i=x+or+%28not+x+and+y%29

starknet_devnet/blueprints/rpc/misc.py
mikiw: fixed

starknet_devnet/blueprints/rpc/misc.py
mikiw: fixed

test/rpc/test_rpc_misc.py
mikiw: fixed

test/rpc/test_rpc_misc.py
mikiw: fixed

starknet_devnet/blueprints/rpc/misc.py
mikiw: True, it's because I'm stupid, fixed.

starknet_devnet/blueprints/rpc/misc.py
mikiw: fixed

starknet_devnet/blueprints/rpc/misc.py
mikiw: fixed

starknet_devnet/blueprints/rpc/misc.py
mikiw: excellent question!

starknet_devnet/blueprints/rpc/misc.py
mikiw: Is not special, it's just an empty value like 0 in ints.

starknet_devnet/blueprints/rpc/misc.py
mikiw: Fixed.

test/rpc/test_data/get_events.py
mikiw: Fixed.

mikiw:
@FabijanC documentation is already updated https://github.com/Shard-Labs/starknet-devnet/blob/master/page/docs/guide/json-rpc-api.md

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +60,47 @@ async def syncing() -> Union[dict, bool]:
...
+        int(state.starknet_wrapper.blocks.get_number_of_blocks())
+        if to_block == "latest"
+        else int(to_block) + 1
+    )
+    for block_number in range(int(from_block), to_block):
+        block = state.starknet_wrapper.blocks.get_by_number(block_number)
+        if block.transaction_receipts != ():
+            events.extend(get_events_from_block(block, address, keys))
+
+    if chunk_size > 0 and continuation_token == "":
```
FabijanC, line 93: This can be simplified as
```python
continuation_token = int(continuation_token)
start_index = continuation_token * chunk_size
events = events[start_index: start_index + chunk_size]
```

`continuation_token` can take a default value of `"0"` instead of `""`

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +60,47 @@ async def syncing() -> Union[dict, bool]:
...
+        events = events[:chunk_size]
+    elif chunk_size > 0 and continuation_token != "":
+        events = (list(islice(events, chunk_size * (int(continuation_token)), None)))[
+            :chunk_size
+        ]
+
+    return RpcEventsResult(
+        events=events,
+        continuation_token=continuation_token,
```
FabijanC, line 102: Should `continuation_token + 1` be returned?

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -27,12 +60,47 @@ async def syncing() -> Union[dict, bool]:
...
+    events = []
+    keys = [] if keys is None else [int(k, 0) for k in keys]
+    to_block = (
+        int(state.starknet_wrapper.blocks.get_number_of_blocks())
+        if to_block == "latest"
+        else int(to_block) + 1
+    )
+    for block_number in range(int(from_block), to_block):
+        block = state.starknet_wrapper.blocks.get_by_number(block_number)
+        if block.transaction_receipts != ():
```
FabijanC, line 90: You can just do `if block.transaction_receipts:` because an empty tuple `()` will take a boolean value of `False`

FabijanC (changes requested):

starknet_devnet/blueprints/rpc/misc.py
mikiw: Fixed.

starknet_devnet/blueprints/rpc/misc.py
mikiw: Yes as we talked about yesterday. Fixed.

starknet_devnet/blueprints/rpc/misc.py
mikiw: fixed

FabijanC (commented):
A conflict needs to be resolved, but other than that LGTM

FabijanC (approved):

PR #313, opened 2022-10-20
Allow RPC calls using block hash or number of latest block

cptartur:
## Usage related changes

- Currently, in many places in the RPC, it is required that the users uses `"latest"` as block hash/number. With this PR it is also possible to use a `block_hash` or `block_number` if they belong to the latest block.
- This PR allows making calls  with block tag `"pending"` block, though it is still treated as a `"latest"` block (up to discussion if this should be included).
- Changed fixtures using  pytest.mark.parametrize usages with `indirect=True`, to parametrized fixtures.

None

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [ ] All tests are passing

### DIFF
--- a/page/docs/guide/json-rpc-api.md
+++ b/page/docs/guide/json-rpc-api.md
@@ -32,7 +32,8 @@ Methods currently not supported:
 
 - `starknet_protocolVersion` - will be removed in a future version of the specification
 
-Methods that require a `block_id` only support ids of the `latest` or `pending` block:
+Methods that require a `block_id` only support ids of the `latest` or `pending` block.
+Please note however, that the `pending` block will be the same block as the `latest`.
 
 ```js
 // Use latest
--- a/page/docs/guide/json-rpc-api.md
+++ b/page/docs/guide/json-rpc-api.md
@@ -31,3 +31,31 @@ Response:
 Methods currently not supported:
 
 - `starknet_protocolVersion` - will be removed in a future version of the specification
+
+Methods that require a `block_id` only support ids of the `latest` or `pending` block:
+
+```json
+// Use latest
+{
+  "block_id": "latest"
+}
+
+// or pending
+{
+  "block_id": "pending"
+}
+
+// or block number
+{
+  "block_id": {
+    "block_number": 1234  // Must be the number of the latest block
+  }
+}
+
+// or block hash
+{
+  "block_id": {
+    "block_hash": "0x1234" // Must be hash of the latest block
+  }
+}
+```
--- a/page/docs/guide/json-rpc-api.md
+++ b/page/docs/guide/json-rpc-api.md
@@ -34,7 +34,7 @@ Methods currently not supported:
 
 Methods that require a `block_id` only support ids of the `latest` or `pending` block:
 
-```json
+```js
 // Use latest
 {
   "block_id": "latest"
--- a/starknet_devnet/blueprints/rpc/utils.py
+++ b/starknet_devnet/blueprints/rpc/utils.py
@@ -53,16 +53,18 @@ def assert_block_id_is_latest_or_pending(block_id: BlockId) -> None:
     Assert block_id is "latest"/"pending" or a block hash or number of "latest"/"pending" block and throw RpcError otherwise
     """
     last_block = state.starknet_wrapper.blocks.get_last_block()
+    print(type(block_id), block_id)
 
-    if isinstance(block_id, BlockHashDict):
-        if block_id["block_hash"] == last_block.block_hash:
-            return
+    if isinstance(block_id, dict):
+        if "block_hash" in block_id:
+            if int(block_id["block_hash"], 16) == last_block.block_hash:
+                return
 
-    if isinstance(block_id, BlockNumberDict):
-        if block_id["block_number"] == last_block.block_number:
-            return
+        if "block_number" in block_id:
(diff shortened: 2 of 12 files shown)


starknet_devnet/blueprints/rpc/utils.py
```diff
@@ -47,15 +51,30 @@ def get_block_by_block_id(block_id: BlockId) -> dict:
...
-    Assert block_id is "latest" and throw RpcError otherwise
+    Assert block_id is "latest"/"pending" or a block hash or number of "latest"/"pending" block and throw RpcError otherwise
     """
-    if block_id != "latest":
-        raise RpcError(
-            code=-1,
-            message="Calls with block_id != 'latest' are not supported currently.",
-        )
+    last_block = state.starknet_wrapper.blocks.get_last_block()
+    print(type(block_id), block_id)
```
FabijanC, line 59: print?

starknet_devnet/blueprints/rpc/utils.py
```diff
@@ -47,15 +51,30 @@ def get_block_by_block_id(block_id: BlockId) -> dict:
...
     """
-    Assert block_id is "latest" and throw RpcError otherwise
+    Assert block_id is "latest"/"pending" or a block hash or number of "latest"/"pending" block and throw RpcError otherwise
     """
-    if block_id != "latest":
-        raise RpcError(
-            code=-1,
-            message="Calls with block_id != 'latest' are not supported currently.",
-        )
+    last_block = state.starknet_wrapper.blocks.get_last_block()
```
FabijanC, line 58: This last block retrieval can be moved inside of the `if isinstance(block_id, dict)` clause

starknet_devnet/blueprints/rpc/utils.py
```diff
@@ -47,15 +51,30 @@ def get_block_by_block_id(block_id: BlockId) -> dict:
...
+            if int(block_id["block_number"]) == last_block.block_number:
+                return
+
+    if isinstance(block_id, str):
+        if block_id in ("latest", "pending"):
+            return
+
+    raise RpcError(
+        code=-1,
+        message="Calls with block_id != 'latest' are not supported currently.",
```
FabijanC, line 76: The error message should also be modified, right? If I may suggest extracting the valid block_ids to a variable and using it for both the actual checking and the message construction.

FabijanC (changes requested):
Thanks for the contribution!

1. Accepted block IDs can be documented in `page/docs/guide/json-rpc-api.md`.
2. Currently only `"latest"` is tested. Shouldn't providing `"pending"` also be tested?

starknet_devnet/blueprints/rpc/utils.py
```diff
@@ -47,15 +51,30 @@ def get_block_by_block_id(block_id: BlockId) -> dict:
...
-    if block_id != "latest":
-        raise RpcError(
-            code=-1,
-            message="Calls with block_id != 'latest' are not supported currently.",
-        )
+    last_block = state.starknet_wrapper.blocks.get_last_block()
+    print(type(block_id), block_id)
+
+    if isinstance(block_id, dict):
+        if "block_hash" in block_id:
```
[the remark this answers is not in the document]
cptartur: I think it's not handled anywhere, I'll include a test for this and a proper check.

cptartur:
1. Will be added
2. I wasn't sure if we want to include "pending", for me it seems like a logical addition since `"pending"` is supported in gateway as far as I know. I'll extend the test to cover pending block.

cptartur:
Addressed the comments. I've also unchecked "tests are passing" because `test/test_postman.py::test_postman` suddenly stopped working. I suspect it's a problem with some local configuration on my end so if tests could be run on your end we could verify that.

page/docs/guide/json-rpc-api.md
```diff
@@ -31,3 +31,31 @@ Response:
 Methods currently not supported:
 
 - `starknet_protocolVersion` - will be removed in a future version of the specification
+
+Methods that require a `block_id` only support ids of the `latest` or `pending` block:
+
+```json
```
FabijanC, line 37: `json` doesn't support comments, so if you could change the language to `js` or `ts`.

page/docs/guide/json-rpc-api.md
```diff
@@ -31,3 +31,31 @@ Response:
 Methods currently not supported:
 
 - `starknet_protocolVersion` - will be removed in a future version of the specification
+
+Methods that require a `block_id` only support ids of the `latest` or `pending` block:
```
FabijanC, line 35: Can you point out that these tags refer to the same block?

test/rpc/test_rpc_call.py
```diff
@@ -52,6 +52,29 @@ def test_call_raises_on_incorrect_contract_address():
     assert ex["error"] == {"code": 20, "message": "Contract not found"}
 
 
+@pytest.mark.usefixtures("run_devnet_in_background", "deploy_info")
+def test_call_raises_on_both_hash_and_number():
+    """
+    Call contract with incorrect address
```
FabijanC, line 58: Docstring not corresponding to test.

FabijanC (changes requested):
Cloned your fork, postman test is passing.

FabijanC (approved):

PR #337, opened 2022-11-10
Fix dumping on exit

mikiw:
## Usage related changes

- Quick fix for double execution of `dump` function.

- `test_dumping_on_exit` updated to cover scenario with call after load

## Checklist:

- [ ] Performed code self-review
- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -107,6 +107,7 @@ def main():
     except KeyboardInterrupt:
         pass
     finally:
+        # Dump only if process is worker (not main)
         if args.dump_on == DumpOn.EXIT and os.getpid() != pid:
             state.dumper.dump()
             sys.exit(0)
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -2,7 +2,9 @@
 A server exposing Starknet functionalities as API endpoints.
 """
 
-import sys, asyncio, os
+import asyncio
+import os
+import sys
 
 from flask import Flask, jsonify
 from flask_cors import CORS
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -102,13 +102,13 @@ def main():
 
     try:
         print(f" * Listening on http://{args.host}:{args.port}/ (Press CTRL+C to quit)")
-        pid = os.getpid()
+        main_pid = os.getpid()
         GunicornServer(app, args).run()
     except KeyboardInterrupt:
         pass
     finally:
         # Dump only if process is worker (not main)
-        if args.dump_on == DumpOn.EXIT and os.getpid() != pid:
+        if args.dump_on == DumpOn.EXIT and os.getpid() != main_pid:
             state.dumper.dump()
             sys.exit(0)
 
--- a/starknet_devnet/server.py
+++ b/starknet_devnet/server.py
@@ -2,9 +2,7 @@
 A server exposing Starknet functionalities as API endpoints.
 """
 
-import sys
-import asyncio
-import os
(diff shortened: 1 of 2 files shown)


mikiw:
This probably is related to introduction of `GunicornServer` and workers:

```
    asyncio.run(state.starknet_wrapper.initialize())

    try:
        print(f" * Listening on http://{args.host}:{args.port}/ (Press CTRL+C to quit)")
        pid = os.getpid()
        GunicornServer(app, args).run()
    except KeyboardInterrupt:
        pass
    finally:
        if args.dump_on == DumpOn.EXIT and os.getpid() != pid:
            state.dumper.dump()
            sys.exit(0)
```

mikiw:
@Solpatium do you have any idea why we have 2 processes running after starting devnet?

image

FabijanC:
Probably due to this line each process has its own "global" state
https://github.com/Shard-Labs/starknet-devnet/blob/40d5b6f75fd80e5adea989e5f45ebb76c5c25cde/starknet_devnet/state.py#L57

starknet_devnet/server.py
```diff
@@ -101,11 +102,12 @@ def main():
 
     try:
         print(f" * Listening on http://{args.host}:{args.port}/ (Press CTRL+C to quit)")
+        pid = os.getpid()
```
FabijanC, line 105: I would call this `main_pid`

starknet_devnet/server.py
```diff
@@ -4,6 +4,7 @@
 
 import sys
 import asyncio
+import os
```
FabijanC, line 7: Can you just sort these three imports alphabetically?

starknet_devnet/server.py
```diff
@@ -101,11 +102,12 @@ def main():
 
     try:
         print(f" * Listening on http://{args.host}:{args.port}/ (Press CTRL+C to quit)")
+        pid = os.getpid()
         GunicornServer(app, args).run()
     except KeyboardInterrupt:
         pass
     finally:
-        if args.dump_on == DumpOn.EXIT:
+        if args.dump_on == DumpOn.EXIT and os.getpid() != pid:
```
FabijanC, line 110: Here I would add a comment saying:
```python
# Dump only if process is worker (not main)
```

FabijanC (changes requested):

starknet_devnet/server.py
mikiw: https://github.com/Shard-Labs/starknet-devnet/pull/337/commits/83d64cdb668ba95ace6aa20eaeebc369f6aae290

starknet_devnet/server.py
mikiw: https://github.com/Shard-Labs/starknet-devnet/pull/337/commits/2c18a620649eacf0521e589c083945bdabe72eff

starknet_devnet/server.py
mikiw: https://github.com/Shard-Labs/starknet-devnet/pull/337/commits/a350c050ab498a10c7d6decde837da54d002c320

mikiw:
Should we use `worker_exit` in gunicorn and always use 1 worker?

Solpatium:
Having more than 1 worker doesn't make sense at the moment as there is no way of sharing the state between them, right?

FabijanC:
> Having more than 1 worker doesn't make sense at the moment as there is no way of sharing the state between them, right?

Yes. If you don't have objections, I'll merge the PR as it currently is.

PR #374, opened 2022-12-29
Mempool blocks generation

mikiw:
## Usage related changes

- Granular block generation added via `--blocks-on-demand` mode and `create_block_on_demand` endpoint. The genesis block will be added normally.

- `generate` method in `DevnetBlocks` class was changed to accept a list of transactions

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/blocks.py
+++ b/starknet_devnet/blocks.py
@@ -124,7 +124,6 @@ class DevnetBlocks:
         block_number = self.get_number_of_blocks()
         timestamp = state.state.block_info.block_timestamp
         signatures = []
-        tx_hashes = []
         internal_transactions = []
 
         if block_number == 0:
@@ -140,7 +139,6 @@ class DevnetBlocks:
             transaction_receipts = tuple(tx.get_execution() for tx in transactions)
             internal_transactions = [tx.internal_tx for tx in transactions]
             signatures = [tx.get_signature() for tx in transactions]
-            tx_hashes = [tx.hash_value for tx in internal_transactions]
 
         if self.lite or is_empty_block:
             block_hash = block_number
@@ -151,7 +149,7 @@ class DevnetBlocks:
                 block_number=block_number,
                 global_state_root=state_root,
                 block_timestamp=timestamp,
-                tx_hashes=tx_hashes,
+                tx_hashes=[tx.hash_value for tx in internal_transactions],
                 tx_signatures=signatures,
                 event_hashes=[],
                 sequencer_address=state.general_config.sequencer_address,
--- a/starknet_devnet/blocks.py
+++ b/starknet_devnet/blocks.py
@@ -2,7 +2,7 @@
 Class for generating and handling blocks
 """
 
-from typing import Dict
+from typing import Dict, List
 
 from starkware.starknet.core.os.block_hash.block_hash import calculate_block_hash
 from starkware.starknet.definitions.error_codes import StarknetErrorCode
@@ -110,7 +110,7 @@ class DevnetBlocks:
 
     async def generate(
         self,
-        transaction: DevnetTransaction,
+        transactions: List[DevnetTransaction],
         state: StarknetState,
         state_update=None,
         is_empty_block=False,
@@ -123,31 +123,36 @@ class DevnetBlocks:
         state_root = DUMMY_STATE_ROOT
         block_number = self.get_number_of_blocks()
         timestamp = state.state.block_info.block_timestamp
+        signatures = []
+        tx_hashes = []
+        internal_transactions = []
+
         if block_number == 0:
             parent_block_hash = 0
         else:
             last_block = await self.get_last_block()
             parent_block_hash = last_block.block_hash
-
+        
         if is_empty_block:
             transaction_receipts = ()
             transactions = []
         else:
-            transaction_receipts = (transaction.get_execution(),)
-            transactions = [transaction.internal_tx]
+            transaction_receipts = (transactions[0].get_execution(),) #TODO: fix this later
+            internal_transactions = [tx.internal_tx for tx in transactions]
+            signatures = [tx.get_signature() for tx in transactions]
+            tx_hashes = [tx.hash_value for tx in internal_transactions]
 
         if self.lite or is_empty_block:
             block_hash = block_number
         else:
-            signature = transaction.get_signature()
             block_hash = await calculate_block_hash(
                 general_config=state.general_config,
                 parent_hash=parent_block_hash,
                 block_number=block_number,
                 global_state_root=state_root,
                 block_timestamp=timestamp,
-                tx_hashes=[transaction.internal_tx.hash_value],
-                tx_signatures=[signature],
+                tx_hashes=tx_hashes,
+                tx_signatures=signatures,
                 event_hashes=[],
                 sequencer_address=state.general_config.sequencer_address,
             )
@@ -156,7 +161,7 @@ class DevnetBlocks:
             block_hash=block_hash,
             block_number=block_number,
             state_root=state_root,
-            transactions=transactions,
+            transactions=internal_transactions,
             timestamp=timestamp,
             transaction_receipts=transaction_receipts,
             status=BlockStatus.ACCEPTED_ON_L2,
--- a/starknet_devnet/blocks.py
+++ b/starknet_devnet/blocks.py
@@ -137,9 +137,7 @@ class DevnetBlocks:
             transaction_receipts = ()
             transactions = []
         else:
-            transaction_receipts = (
-                transactions[0].get_execution(),
-            )  # TODO: fix this later
+            transaction_receipts = tuple(tx.get_execution() for tx in transactions)
             internal_transactions = [tx.internal_tx for tx in transactions]
             signatures = [tx.get_signature() for tx in transactions]
             tx_hashes = [tx.hash_value for tx in internal_transactions]
--- a/starknet_devnet/blocks.py
+++ b/starknet_devnet/blocks.py
@@ -132,12 +132,14 @@ class DevnetBlocks:
         else:
             last_block = await self.get_last_block()
             parent_block_hash = last_block.block_hash
-        
+
         if is_empty_block:
             transaction_receipts = ()
             transactions = []
         else:
-            transaction_receipts = (transactions[0].get_execution(),) #TODO: fix this later
+            transaction_receipts = (
+                transactions[0].get_execution(),
+            )  # TODO: fix this later
             internal_transactions = [tx.internal_tx for tx in transactions]
             signatures = [tx.get_signature() for tx in transactions]
             tx_hashes = [tx.hash_value for tx in internal_transactions]
--- a/starknet_devnet/blocks.py
+++ b/starknet_devnet/blocks.py
@@ -123,8 +123,9 @@ class DevnetBlocks:
         state_root = DUMMY_STATE_ROOT
         block_number = self.get_number_of_blocks()
         timestamp = state.state.block_info.block_timestamp
-        signatures = []
-        internal_transactions = []
+        signatures = [tx.get_signature() for tx in transactions or []]
+        internal_transactions = [tx.internal_tx for tx in transactions or []]
+        transaction_receipts = tuple(tx.get_execution() for tx in transactions or ())
 
         if block_number == 0:
             parent_block_hash = 0
@@ -132,16 +133,9 @@ class DevnetBlocks:
             last_block = await self.get_last_block()
             parent_block_hash = last_block.block_hash
 
-        if is_empty_block:
-            transaction_receipts = ()
-            transactions = []
-        else:
-            transaction_receipts = tuple(tx.get_execution() for tx in transactions)
-            internal_transactions = [tx.internal_tx for tx in transactions]
-            signatures = [tx.get_signature() for tx in transactions]
-
         if self.lite or is_empty_block:
             block_hash = block_number
+            transactions = []
         else:
             block_hash = await calculate_block_hash(
                 general_config=state.general_config,
--- a/starknet_devnet/blueprints/base.py
+++ b/starknet_devnet/blueprints/base.py
@@ -194,6 +194,13 @@ async def create_block():
     block = await state.starknet_wrapper.create_empty_block()
     return Response(block.dumps(), status=200, mimetype="application/json")
 
+@base.route("/create_block_on_demand", methods=["POST"])
+async def create_block_on_demand():
+    """Create block on demand"""
+    block = await state.starknet_wrapper.store_pending_transactions()
+    print(block.transactions)
+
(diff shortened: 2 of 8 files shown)


mikiw:
In the task it was mentioned, "Thus, all transactions within will always run in a single block unless it's specified not too."

So we want to specify how many transactions will be included in a block like `max-transactions-per-block` it can be 1 or 100?

mikiw:
I'm not sure why in my tests hashes are non-deterministic. I'm using magical 42 seed.

mikiw:
I had linter error on my local:
```
************* Module starknet_devnet.cairo_rs_py_patch
starknet_devnet/cairo_rs_py_patch.py:13:0: E0401: Unable to import 'cairo_rs_py' (import-error)
```
but seems that there is no issue on circle ci

mikiw:
Discussed on the call with @FabijanC:
```
        """
        Stores the provided transactions in new block.
        """
        # Fabijan I'm not sure about this... Let's discuss it during PR.
        # Update this according to _store_transaction() method?
        return await self.blocks.generate(
            transactions,
            self.get_state(),
        )
```

mikiw:
Ariel suggested that Mempool is a better name for that feature and it makes sense not to confuse it with pending transactions.

test/test_blocks_on_demand.py
```diff
@@ -0,0 +1,106 @@
...
+
+
+@devnet_in_background(
+    *[
+        *PREDEPLOY_ACCOUNT_CLI_ARGS,
+        "--blocks-on-demand",
+    ]
+)
+def test_blocks_on_demand_invoke():
+    """Test deploy in blocks on demand mode"""
```
FabijanC, line 28: Could you write it as `blocks-on-demand` (with hyphens) so it's easier to understand.

test/test_blocks_on_demand.py
```diff
@@ -0,0 +1,106 @@
...
+        call(
+            function="get_balance",
+            address=deploy_info["address"],
+            abi_path=ABI_PATH,
+        )
+    except ReturnCodeAssertionError:
+        should_fail_on_contract_call = True
+    assert should_fail_on_contract_call is True
+
+    invoke(
```
FabijanC, line 45: If the deploy tx is not considered either PENDING or ACCEPTED_ON_L2 at this point, does it make sense to be able to invoke? What is the behavior on alpha-goerli for this scenario?

starknet_devnet/blueprints/base.py
```diff
@@ -195,6 +195,14 @@ async def create_block():
     return Response(block.dumps(), status=200, mimetype="application/json")
 
 
+@base.route("/create_block_on_demand", methods=["POST"])
+async def create_block_on_demand():
+    """Create block on demand"""
+    block = await state.starknet_wrapper.store_mempool_transactions()
```
FabijanC, line 201: What happens if no txs in the mempool?

starknet_devnet/blocks.py
```diff
@@ -123,6 +123,9 @@ async def generate(
         state_root = DUMMY_STATE_ROOT
         block_number = self.get_number_of_blocks()
         timestamp = state.state.block_info.block_timestamp
+        signatures = []
```
FabijanC, line 126: Why add these new arguments? Can't they be extracted from `DevnetTransaction`?

starknet_devnet/starknet_wrapper.py
```diff
@@ -105,6 +106,8 @@ def __init__(self, config: DevnetConfig):
         self.fee_token = FeeToken(self)
         self.accounts = Accounts(self)
         self.__udc = UDC(self)
+        self.mempool = []
```
FabijanC, line 109: If these properties are only meant to be accessed from inside starknet_wrapper, they can be prefixed with a double underscore (`__`).

FabijanC (changes requested):
`get_state_update` response generation didn't have to be modified?

starknet_devnet/starknet_wrapper.py
```diff
@@ -450,7 +481,11 @@ async def invoke(self, external_tx: InvokeFunction):
     async def call(self, transaction: CallFunction):
         """Perform call according to specifications in `transaction`."""
 
-        state_copy = self.get_state().copy()
+        if self.config.blocks_on_demand and self.mempool_state is not None:
+            state_copy = self.mempool_state
```
FabijanC, line 485: Shouldn't this be `self.mempool_state.copy()`? E.g. if a call was made twice in a row, wouldn't it produce different results since `mempool_state` was actually altered?

starknet_devnet/starknet_wrapper.py
```diff
@@ -348,12 +376,15 @@ async def __aexit__(
...
                 )
 
-                await self.starknet_wrapper._store_transaction(
-                    transaction=transaction,
-                    state_update=state_update,
-                    error_message=error_message,
-                    tx_hash=tx_hash,
-                )
+                if self.starknet_wrapper.config.blocks_on_demand:
+                    self.starknet_wrapper.mempool.append(transaction)
```
FabijanC, line 380: When and where is mempool cleared?

starknet_devnet/starknet_wrapper.py
```diff
@@ -450,7 +481,11 @@ async def invoke(self, external_tx: InvokeFunction):
     async def call(self, transaction: CallFunction):
```
FabijanC, line 481: Is `call` really the only place that needs to be modified to use `mempool_state`?

FabijanC (changes requested):
1. Should now received transactions be marked as `RECEIVED`? And only when the block is created they should become `ACCEPTED_ON_L2`?

2. To reduce the number of ifs and simplify the logic, couldn't the mempool state always be used for queries? Instead of having to check whether `blocks-on-demand` is turned on?

test/test_blocks_on_demand.py
```diff
@@ -0,0 +1,106 @@
...
+    CONTRACT_PATH,
+    PREDEPLOY_ACCOUNT_CLI_ARGS,
+    PREDEPLOYED_ACCOUNT_ADDRESS,
+    PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+)
+from .util import ReturnCodeAssertionError, call, deploy, devnet_in_background
+
+
+@devnet_in_background(
+    *[
```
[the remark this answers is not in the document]
mikiw: true, I missed that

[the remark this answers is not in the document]
mikiw: Nice, thanks fixed in 3509d7fd7f7794b2ce86a7bcf91546e5aac3b7f4

test/test_blocks_on_demand.py
```diff
@@ -0,0 +1,106 @@
...
+        "--blocks-on-demand",
+    ]
+)
+def test_blocks_on_demand_invoke():
+    """Test deploy in blocks on demand mode"""
+    latest_block = gateway_call("get_block", blockNumber="latest")
+    genesis_block_number = latest_block["block_number"]
+    assert genesis_block_number == 0
+
+    deploy_info = deploy(CONTRACT_PATH, inputs=["0"])
```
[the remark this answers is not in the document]
mikiw: NOT_RECEIVED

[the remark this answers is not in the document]
FabijanC: We would want it to be RECEIVED, like on alpha-goerli, right?

starknet_devnet/blueprints/base.py
mikiw: An empty block will be generated.

FabijanC: Ok, this should be documented then

starknet_devnet/starknet_wrapper.py
mikiw: but is also used in `TransactionHandler` class

FabijanC: It is sort of a private class, but if the linter complains, let's leave it as it is.

test/test_blocks_on_demand.py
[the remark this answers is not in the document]
mikiw: Fixed here b67e6260185b946655770e078bb150b30267cd01

starknet_devnet/blueprints/feeder_gateway.py
```diff
@@ -210,6 +211,18 @@ async def get_transaction_status():
...
+    # In case of blocks-on-demand feature tx_status should be received
+    if (
+        state.starknet_wrapper.config.blocks_on_demand
+        and tx_status["tx_status"] == TransactionStatus.NOT_RECEIVED.name
+        and any(
+            hex(transaction.transaction_hash) == transaction_hash
+            for transaction in state.starknet_wrapper.mempool
+        )
+    ):
+        tx_status["tx_status"] = TransactionStatus.RECEIVED.name
```
FabijanC, line 224: TBH this looks hacky. Why can't we just store them as RECEIVED, and then during block creation just change the status of each tx to ACCEPTED_ON_L2

starknet_devnet/starknet_wrapper.py
mikiw: yes I tried to add this but tests were failing and I realized that it's also used in `TransactionHandler`

mikiw: Fixed 7d8ed9a605d98923d6c801e996bcecf437bd17b5

test/test_blocks_on_demand.py
```diff
@@ -0,0 +1,106 @@
...
+    should_fail_on_contract_call = False
+    try:
+        call(
+            function="get_balance",
+            address=deploy_info["address"],
+            abi_path=ABI_PATH,
+        )
+    except ReturnCodeAssertionError:
+        should_fail_on_contract_call = True
```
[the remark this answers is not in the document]
mikiw: Fixed in 2e7b7f709ce804f3540447e1b2e5a3e10bd10789

test/test_blocks_on_demand.py
```diff
@@ -0,0 +1,106 @@
...
+
+    deploy_info = deploy(CONTRACT_PATH, inputs=["0"])
+    assert_tx_status(deploy_info["tx_hash"], "RECEIVED")
+
+    try:
+        call(
+            function="get_balance",
+            address=deploy_info["address"],
+            abi_path=ABI_PATH,
+        )
```
FabijanC, line 42: After `call`, this must `raise` a RuntimeError saying that the previous line should have raised. It will only be raised if `call` itself did not raise. Like here (it seems that pytest has `pytest.fail()` which could be used with an appropriate message):
https://github.com/Shard-Labs/starknet-hardhat-example/blob/7d3381600caed8e29c2e8de1e2cda3fb04aa78d5/test/sample-test.ts#L67

starknet_devnet/blueprints/base.py
mikiw: fixed in dd8dc8d8b07198cb21e48665772b79891aa101d9

FabijanC: Do users know what a mempool is? Also the genesis block info looks out of place: it looks like it's related to the previous sentence, but I don't think it should be. Does it even have to be mentioned here?

starknet_devnet/starknet_wrapper.py
mikiw: Good that you spotted that bug, it should be cleared in `store_mempool_transactions()` but it's not... I can't explain why I forgot about this. I'll also add this to tests.

starknet_devnet/blueprints/feeder_gateway.py
mikiw: Ok, let's discuss it later on a call.

starknet_devnet/starknet_wrapper.py
mikiw: Fixed in 7258b975bc3d67cd3b6ef21968beb844c927d03c

test/test_blocks_on_demand.py
mikiw: Fixed in 120646c326153f6908c609b1b257bba40b32abae

starknet_devnet/blueprints/base.py
mikiw: Mempool is a common term in blockchain terminology, they can google that also.

I think that it's worth mentioning so the user, doesn't need to execute `/create_block_on_demand` immediately after devnet starts.

FabijanC: > Mempool is a common term in blockchain terminology, they can google that also.

Ok, makes sense.

> I think that it's worth mentioning so the user, doesn't need to execute `/create_block_on_demand` immediately after devnet starts.

Sure that piece of information should be in the docs, but should it be in the very paragraph where it is right now. When I see this:
> In case of no transactions in mempool new empty block will be generated. The genesis block will be generated normally.

I consider the genesis block generation to be related to the mempool being empty. We are not trying to say that, right?

mikiw: Different blockchains generate genesis blocks differently, sometimes genesis block contains transactions or some state at the beginning. I just wanted to point out that it this case with this flag genesis block will be generated and will be empty.

FabijanC: ok, resolving this conversation

test/test_blocks_on_demand.py
mikiw: It took some time but I tested it and when the transaction is it's `Received` you can't `invoke, there is a `StarknetErrorCode.UNINITIALIZED_CONTRACT` error, I couldn't test the `PENDING` state since it's an instantly `ACCEPTED_ON_L2` after `Received`

When it's `ACCEPTED_ON_L2` you can `invoke` it.

starknet_devnet/starknet_wrapper.py
mikiw: Let's discuss it later on a call since the user can interact with contract only when `deploy` is called and it's accepted on l2.

starknet_devnet/blocks.py
mikiw: I came up with something like this ce5e1bfd8e2033b22435ef62429cee21d16bad09

mikiw:
> 2. To reduce the number of ifs and simplify the logic, couldn't the mempool state always be used for queries? Instead of having to check whether `blocks-on-demand` is turned on?

But we would need to populate this line `self.__mempool_state = self.get_state().copy()` in a lot of places in code. Didn't we?

starknet_devnet/starknet_wrapper.py
mikiw: but mempool state is already a copy of state:
```
        # Update mempool state before block generation
        self.__mempool_state = self.get_state().copy()
```
so now it will be referenced to variable `state_copy` and a call will be made.

FabijanC:
Superseded by #389. Closing.

PR #386, opened 2023-01-20
Get events filter bugfix

mikiw:
## Usage related changes

- bugfix for `getEvents` rpc

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing

### DIFF
--- a/starknet_devnet/blueprints/rpc/misc.py
+++ b/starknet_devnet/blueprints/rpc/misc.py
@@ -92,11 +92,7 @@ async def get_events(filter) -> RpcEventsResult:
     # Optional parameters
     address = try_parse_from_filter(filter, "address")
     keys = try_parse_from_filter(filter, "keys")
-    continuation_token = try_parse_from_filter(filter, "continuation_token")
-
-    # Set continuation_token to default value "0" when it's not set
-    if continuation_token is None:
-        continuation_token = "0"
+    continuation_token = filter.get("continuation_token", "0")
 
     events = []
     keys = [] if keys is None else [int(k, 0) for k in keys]
--- a/starknet_devnet/blueprints/rpc/misc.py
+++ b/starknet_devnet/blueprints/rpc/misc.py
@@ -84,7 +84,7 @@ async def get_events(filter) -> RpcEventsResult:
     except ValueError as ex:
         raise RpcError(
             code=RpcErrorCode.INVALID_PARAMS.value,
-            message="unknown variant `0`, expected one of `block_number`, `block_hash`, `latest`, `pending`",
+            message=f"invalid chunk_size: '{filter.get('chunk_size')}'",
         ) from ex
 
     # Optional parameters
--- a/starknet_devnet/blueprints/rpc/misc.py
+++ b/starknet_devnet/blueprints/rpc/misc.py
@@ -73,7 +73,7 @@ def try_parse_from_filter(filter_data, name):
         return None
 
 
-async def get_events(filter_data) -> str:
+async def get_events(filter_data) -> RpcEventsResult:
     """
     Returns all events matching the given filters.
 
--- a/starknet_devnet/blueprints/rpc/misc.py
+++ b/starknet_devnet/blueprints/rpc/misc.py
@@ -63,14 +63,19 @@ async def syncing() -> Union[dict, bool]:
     return False
 
 
+def tryParseFromFilter(filter_data, name):
+    """
+    Try to parse data from filter data. Return None if there is no data for a given name.
+    """
+    try:
+        return filter_data[name]
+    except KeyError:
+        return None
+
+
 # pylint: disable=too-many-arguments
 async def get_events(
-    from_block: BlockId,
-    to_block: BlockId,
-    chunk_size: int,
-    address: Address = None,
-    keys: List[Address] = None,
-    continuation_token: str = "0",
+    filter,
 ) -> str:
     """
     Returns all events matching the given filters.
@@ -82,6 +87,21 @@ async def get_events(
     and chunk it later which is not an optimal solution.
     """
     events = []
+    
+    # Required parameters
+    from_block = tryParseFromFilter(filter, "from_block")
+    to_block = tryParseFromFilter(filter, "to_block")
+    chunk_size = int(tryParseFromFilter(filter, "chunk_size"))
+
+    # Optional parameters
+    address = tryParseFromFilter(filter, "address")
+    keys = tryParseFromFilter(filter, "keys")
+    continuation_token = tryParseFromFilter(filter, "continuation_token")
+
+    # Set continuation_token to default value "0" when it's not set
+    if continuation_token is None:
+        continuation_token = "0"
+
     keys = [] if keys is None else [int(k, 0) for k in keys]
     to_block = (
         int(state.starknet_wrapper.blocks.get_number_of_blocks())
@@ -93,12 +113,12 @@ async def get_events(
         if block.transaction_receipts:
             events.extend(get_events_from_block(block, address, keys))
 
-    # chunking
+    # Chunking
     continuation_token = int(continuation_token)
     start_index = continuation_token * chunk_size
     events = events[start_index : start_index + chunk_size]
 
-    # continuation_token should be increased only if events are not empty
+    # Continuation_token should be increased only if events are not empty
     if events:
         continuation_token = continuation_token + 1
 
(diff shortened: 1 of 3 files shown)


test/rpc/test_rpc_misc.py
```diff
@@ -165,14 +165,13 @@ def test_get_events(input_data, expected_data):
...
-    resp = rpc_call("starknet_getEvents", params=input_data)
+    resp = rpc_call("starknet_getEvents", params=input_data["params"])
     assert len(expected_data) == len(resp["result"]["events"])
     for i, data in enumerate(expected_data):
         assert str(resp["result"]["events"][i]["data"]) == str(data)
 
-    if "continuation_token" in input_data:
-        expected_continuation_token = int(input_data["continuation_token"])
-
+    if "continuation_token" in input_data["params"]["filter"]:
```
FabijanC, line 173: What if the continuation_token is not present but it is expected to be?

test/rpc/test_data/get_events.py
```diff
@@ -9,82 +9,161 @@
...
 BLOCK_FROM_0_TO_LATEST_WITH_ADDRESS_AND_KEYS = {
-    "from_block": "0",
-    "to_block": "latest",
-    "address": EXPECTED_FEE_TOKEN_ADDRESS,
-    "keys": [
-        "0x99cd8bde557814842a3121e8ddfd433a539b8c9f14bf31ebf108d12e6196e9",
-        "0x3db3da4221c078e78bd987e54e1cc24570d89a7002cefa33e548d6c72c73f9d",
-    ],
-    "chunk_size": 10,
+    "jsonrpc": "2.0",
```
FabijanC, line 152: All this `BLOCK_FROM_...` could be constructed using a common function which would insert properties like `jsonrpc`, `method`, `id` since the only thing different is the `filter` property of `params`. This refactor can be done in a separate PR.

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -63,15 +63,18 @@ async def syncing() -> Union[dict, bool]:
...
-# pylint: disable=too-many-arguments
-async def get_events(
-    from_block: BlockId,
-    to_block: BlockId,
-    chunk_size: int,
-    address: Address = None,
-    keys: List[Address] = None,
-    continuation_token: str = "0",
-) -> str:
+def try_parse_from_filter(filter_data, name):
```
FabijanC, line 66: The pythonic way to do this is very simple:
```python
filter_data.get(key)
```

This defaults to returning `None` if the key is not present. In fact it can return any default value using the second argument, but the implicit default is `None`. Docs: [link](https://python-reference.readthedocs.io/en/latest/docs/dict/get.html)

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -81,6 +84,20 @@ async def get_events(
...
+    to_block = try_parse_from_filter(filter, "to_block")
+    chunk_size = int(try_parse_from_filter(filter, "chunk_size"))
+
+    # Optional parameters
+    address = try_parse_from_filter(filter, "address")
+    keys = try_parse_from_filter(filter, "keys")
+    continuation_token = try_parse_from_filter(filter, "continuation_token")
+
+    # Set continuation_token to default value "0" when it's not set
+    if continuation_token is None:
```
FabijanC, line 98: This could be
```python
filter.get("continuation_token", "0")
```

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -63,15 +63,18 @@ async def syncing() -> Union[dict, bool]:
...
+    """
+    Try to parse data from filter data. Return None if there is no data for a given name.
+    """
+    try:
+        return filter_data[name]
+    except KeyError:
+        return None
+
+# pylint: disable=redefined-builtin
+# filter name is determined by current RPC implementation and starknet specification
```
FabijanC, line 76: Why not just call the property `filter_data`, that way there is no need to redefine the builtin.

FabijanC (changes requested):

test/rpc/test_rpc_misc.py
mikiw: Let's say someone will remove `continuation_token` for these test calls. Then `continuation_token` will be set to `0` as the default value in `get_events` function and firsts events will be returned and one test will fail because `expected_data` will be different. 

image
Just tested it

FabijanC: I slightly understand (not completely), but I'll believe you on this one. Marking as resolved.

starknet_devnet/blueprints/rpc/misc.py
mikiw: I tried that but it will lead to problems in this code:

```

rpc = Blueprint("rpc", __name__, url_prefix="/rpc")


@rpc.route("", methods=["POST"])
async def base_route():
    """
    Base route for RPC calls
    """

    message_id = None
    try:
        method, params, message_id = parse_body(request.json)
        result = await (
            method(*params) if isinstance(params, list) else method(**params)
        )
        if inspect.iscoroutinefunction(result):
            result = await result
    except TypeError as type_error:
        return rpc_error(message_id=message_id, code=22, message=str(type_error))
    except RpcError as error:
        return rpc_error(message_id=message_id, code=error.code, message=error.message)

    return rpc_response(message_id=message_id, content=result)


def parse_body(body: dict) -> Tuple[Callable, Union[List, dict], int]:
    """
    Parse rpc call body to function name, params and message id
    """
    try:
        method_name = body["method"].replace("starknet_", "")
        params: Union[List, dict] = body.get("params") or {}
        message_id = body["id"]
    except RuntimeError as error:
        raise RpcError(
            code=RpcErrorCode.INVALID_REQUEST.value, message="Invalid request"
        ) from error

    if method_name not in methods:
        raise RpcError(
            code=RpcErrorCode.METHOD_NOT_FOUND.value, message="Method not found"
        )

    if not isinstance(params, (List, Dict)):
        raise RpcError(code=RpcErrorCode.INVALID_PARAMS.value, message="Invalid params")

    return methods[method_name], params, message_id
```

FabijanC: Ah right, ok! We can leave as it is with the suppression. Marking as resolved.

starknet_devnet/blueprints/rpc/misc.py
mikiw: Nice! I was looking exactly for something like this. Fixed in b34e920d6f114743464b8a6eb0f77db9c558a628

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -81,6 +75,16 @@ async def get_events(
     This is why we need to iterate block by block, take all events,
     and chunk it later which is not an optimal solution.
     """
+    # Required parameters
+    from_block = filter.get("from_block")
+    to_block = filter.get("to_block")
+    chunk_size = int(filter.get("chunk_size"))
```
FabijanC, line 81: what happens if `chunk_size` is not convertable to `int`? Does it lead to a 500 error?

FabijanC (changes requested):

test/rpc/test_data/get_events.py
mikiw: I was thinking about this but I remember that I was doing something similar 1 or 2 months ago and we decided that more clean way is not to use something like this.

I'll do it in this PR. It's faster than creating new task etc.

starknet_devnet/blueprints/rpc/misc.py
mikiw: Yes it will lead to 500

FabijanC: But we don't really want that, do we?

mikiw: It's a malformed request. Do you want to return some response with information?

FabijanC: Indicating exactly what caused it (param name) would be good

mikiw: I'll mimic the alpha goerli response

mikiw: Fixed in e9e168cf2b327d9b1a80bf0380864fd6213e3724

PR #387, opened 2023-01-20
Add RPC schema validation

### LINKED ISSUE #379: Add RPC responses and requests validation against json schema
Add RPC responses a request validation again json schema. Library `jsonschema` that is already present in package dependencies and used in a couple places, could be used to do it quite easily.

cptartur:
## Usage related changes

Closes #379 

- Added RPC request and response validation agains the RPC schema
- Added CLI options to disable both request and response validations 

Since 0.2.1 RPC spec had some problems, I've bundled both read and write api with some of my modifications (changed oneOf to anyOf in couple of places, defined missing `required` in write api, etc.)

Note: I didn't modify `required` in most parts of the read specification. Until we migrate to new RPC spec (0.3.0 or different), the validation might still not catch some errors, namely some missing parameters in responses if the specification uses inline schema instead of `$ref`.

- Changed dependencies of the package

## Checklist:

- [x] Documented the changes
- [x] Updated the tests
- [x] All tests are passing


pyproject.toml
```diff
@@ -22,17 +22,17 @@ marshmallow = "~3.17.0"
 typing-extensions = "~4.3.0"
 gunicorn = "~20.1.0"
 marshmallow-dataclass = "~8.4"
+jsonschema = "~4.17.0"
```
cptartur, line 25: It is now used in the app itself, so it must be a standard dependency.

pyproject.toml
```diff
@@ -22,17 +22,17 @@ marshmallow = "~3.17.0"
...
 pylint = "~2.12.2"
-web3 = "~5.28.0"
 psutil = "~5.9.1"
-jsonschema = "~3.2.0"
 pytest-xdist = "~2.5.0"
 pylint-quotes = "~0.2.3"
 black = "~22.6"
 requests = "~2.28"
 isort = "^5.10.1"
+web3 = {version = "6.0.0b4", extras = ["ipfs", "tester"]}
```
cptartur, line 35: This is controversial. I bumped this to a new version, because ~5.28.0 would limit us to using jsonschema `<4.0.0` (and also it caused problems when using devnet in starknet.py 😅 )

starknet_devnet/blueprints/rpc/misc.py
```diff
@@ -64,6 +67,9 @@ async def syncing() -> Union[dict, bool]:
 
 
 # pylint: disable=too-many-arguments
+# Events response does not currently conform to RPC specs
+# and will need fixing before validation is added
+# @validate_schema("getEvents")
```
cptartur, line 72: I discovered this while working on schema validation, but I'd prefer this is addressed in the separate issue.

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,320 @@
...
+    assert error["code"] == RpcErrorCode.INVALID_PARAMS.value
+
+
+@pytest.mark.usefixtures("run_devnet_in_background")
+@pytest.mark.parametrize(
+    "run_devnet_in_background",
+    [["--disable-rpc-request-validation"]],
+    indirect=True,
+)
+def test_schema_does_not_raise_on_disabled_request_validation():
```
cptartur, line 181: I wanted to also add a test for `--disable-rpc-response-validation` but I don't know how to approach that: It'd require somehow mocking the route so it returns incorrect value. 

I did some manual testing and it seems to be working.

starknet_devnet/blueprints/rpc/structures/responses.py
```diff
@@ -200,8 +195,10 @@ async def txn_type() -> RpcTxnType:
         "transaction_hash": rpc_felt(txr.transaction_hash),
         "actual_fee": rpc_felt(txr.actual_fee or 0),
         "status": status(),
-        "block_hash": rpc_felt(txr.block_hash) if txr.block_hash is not None else None,
-        "block_number": txr.block_number or None,
+        "block_hash": rpc_felt(txr.block_hash)
+        if txr.block_hash is not None
+        else rpc_felt(0),
+        "block_number": txr.block_number if txr.block_number is not None else 0,
```
cptartur, line 201: Spec doesn't allow nulls for block_hash and number even for rejected transactions. 0 seemed like reasonable default but I'm unsure.

cptartur (commented):
I didn't tick that documentation has been added, because apart from adding docstrings to introduced methods, no modification was created.

Should some info about this validation be included in the docs itself?

starknet_devnet/blueprints/rpc/schema.py
```diff
@@ -0,0 +1,237 @@
...
+    """
+    Wrapper for ValidationError raised during request validation
+    """
+
+    def __init__(self, err: ValidationError):
+        super().__init__("Failed to validate schema for params.")
+        self.validation_error = err
+
+    def __str__(self):
+        # fmt: off
```
FabijanC, line 186: Here and in other places, instead of toggling formatting off and on, we can simply rely on using triple quotes:
```python
f"""Got invalid value for parameter: "{self.validation_error.message}""""
```

starknet_devnet/blueprints/rpc/structures/responses.py
```diff
@@ -176,15 +175,11 @@ def events() -> List[Event]:
...
-            return "UNKNOWN"
-
-        mapping = {
-            TransactionStatus.NOT_RECEIVED: "UNKNOWN",
+    def status() -> TxnStatus:
+        mapping: Dict[TransactionStatus, TxnStatus] = {
             TransactionStatus.ACCEPTED_ON_L2: "ACCEPTED_ON_L2",
             TransactionStatus.ACCEPTED_ON_L1: "ACCEPTED_ON_L1",
-            TransactionStatus.RECEIVED: "RECEIVED",
+            TransactionStatus.RECEIVED: "PENDING",
```
FabijanC, line 182: Just to check, this is intended to map received to pending?

starknet_devnet/blueprints/rpc/structures/responses.py
FabijanC: When would `txr.block_hash` even be `None`?

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,320 @@
...
+    """
+    resp = rpc_call("starknet_call", params=params)
+
+    assert "error" in resp
+    error = resp["error"]
+
+    assert error["code"] == RpcErrorCode.INVALID_PARAMS.value
+
+
+@pytest.mark.usefixtures("run_devnet_in_background")
```
FabijanC, line 175: This can be rewritten as:
```python
@devnet_in_background("--disable-rpc-request-validation")
def test_...
```

test/rpc/test_rpc_schema.py
FabijanC: First I'd like to hear why we need that flag (asked that in another PR comment)

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,320 @@
...
+                "contract_address": 1234,
+                "entry_point_selector": -1,
+                "calldata": ["a", "b", "c"],
+            },
+        },
+    )
+
+    # Error will be raised when trying to execute function, but it shouldn't be the INVALID_PARAMS error
+    error = resp["error"]
+    assert error["code"] != RpcErrorCode.INVALID_PARAMS.value
```
FabijanC, line 199: Looks shaky; we can't predict the error type?

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,320 @@
...
+def test_schema_does_not_raise_on_correct_kwargs(params):
+    """
+    Test kwargs validation allows valid requests
+    """
+
+    resp = rpc_call("starknet_call", params=params)
+
+    # Error will be raised because address is correctly formatted but incorrect
+    error = resp["error"]
+    assert all(error["code"] != code.value for code in RpcErrorCode)
```
FabijanC, line 251: Shouldn't it return some expected error? Something that we can assert? If I understand this correctly, we are just asserting the received error code is not one of the predefined RPC error codes?

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,320 @@
...
+                "entry_point_selector": rpc_felt(get_selector_from_name("get_balance")),
+                "calldata": [],
+            },
+            "latest",
+        ],
+    )
+
+    # Error will be raised because address is correctly formatted but incorrect
+    error = resp["error"]
+    assert all(error["code"] != code.value for code in RpcErrorCode)
```
FabijanC, line 274: Same as above

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,320 @@
...
+                        }
+                    }
+                },
+            },
+        }
+
+        # Testing private methods is not ideal, but I have no other idea of testing this logic
+        # because mocking doesn't work with devnet running in background subprocess
+        params = {"key": "0x01"}
+        _assert_valid_rpc_request(**params, method_name="starknet_method")
```
FabijanC, line 320: I fail to understand what exactly is tested here and why this test is necessary.

FabijanC (changes requested):
I don't understand why the testing of validation is necessary? My understanding was that the validation is here to provide another layer of security regarding the input/output format.

Other than testing, the observations I left are very minor.

Also, there are conflicts with the base branch, nothing major.

starknet_devnet/blueprints/rpc/schema.py
```diff
@@ -0,0 +1,237 @@
...
+            schema["required"] = list(schema["properties"].keys())
+
+    return methods, schemas
+
+
+def _extract_methods(specs_json: Dict[str, Any]) -> Dict[str, Any]:
+    return {method["name"]: method for method in specs_json["methods"]}
+
+
+def _response_schema_for_method(name: str) -> Dict[str, Any]:
```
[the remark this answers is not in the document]
cptartur: It was intentional, but not sure if my logic makes sense: RPC endpoint has one return type so `response_schema` and it has multiple parameters each with it's own schema so `request_schemas`. But I can unify this if you think that's better.

starknet_devnet/devnet_config.py
```diff
@@ -296,6 +296,16 @@ def parse_args(raw_args: List[str]):
         default=StarknetChainId.TESTNET,
         help=f"Specify the chain id as string: {{{CHAIN_IDS}}}",
     )
+    parser.add_argument(
+        "--disable-rpc-request-validation",
+        action="store_true",
+        help="Disable requests schema validation for RPC endpoints",
```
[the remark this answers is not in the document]
cptartur: The usacase I had in mind is if for whatever reason validation in the devnet rpc had some bugs and was not accepting the inputs that are correct according to specs or the rpc tries to return malformated responses and fails, a user might want to temporarily disable this to be able to continue using the rpc while waiting for the fix.

I think it makes more sense with request validation, response validation might be a bit of a stretch.

starknet_devnet/blueprints/rpc/structures/responses.py
cptartur: It is. RPC defines transaction status as following

```json
"TXN_STATUS": {
    "type": "string",
    "enum": [
        "PENDING",
        "ACCEPTED_ON_L2",
        "ACCEPTED_ON_L1",
        "REJECTED"
    ],
    "description": "The status of the transaction"
}
```

With `RECEIVED` not existing as a status at all. I though `PENDING` would be the logical mapping in this case.

test/rpc/test_rpc_schema.py
cptartur: Yes, it is meant to validate if the error returned was not one of the "internal" RPC errors i.e. not defined in the spec itself.

starknet_devnet/blueprints/rpc/structures/responses.py
cptartur: Not sure honestly, but in the cairo-lang `TransactionInBlockInfo` both block hash and number are typed as `Optional[int]`.

test/rpc/test_rpc_schema.py
cptartur: `params` of the most RPC methods, alongside with the schema, have a `"required": true` field (no method exist currently with required being false). My schema validation is supposed to support not required params anyways.

This test is meant to verify if this logic does work, you can see `is_required` keys in schemas used in mock. In non-mocked situation these should be added by the schema parser based on value of `"required"` key in the schema.

starknet_devnet/blueprints/rpc/schema.py
```diff
@@ -0,0 +1,237 @@
...
+                    # fmt: off
+                    raise ValidationError(f"Missing keyword argument \"{name}\".")
+                    # fmt: on
+                continue
+
+            value = kwargs[name]
+            validate(value, schema)
+        return
+
+    if length := len(schemas) != 0:
```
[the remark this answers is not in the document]
cptartur: Oh, right. Will fix that.

cptartur:
> I don't understand why the testing of validation is necessary? My understanding was that the validation is here to provide another layer of security regarding the input/output format.

I think that incorrectly working validation can be more harmful than good: If it were to raise errors on correctly formatted inputs or outputs, it would make devnet's rpc unusable. Also, it accepting incorrect inputs or outputs (which it still might do, but tried to limit the cases of this with these tests), will reduce the chances of us introducing new bugs to the rpc because "it is being validated anyway".

test/rpc/test_rpc_schema.py
cptartur: Changed so it checks if error code is one of the codes defined as `starknet_call` possible errors. Don't want to hardcode specific error as I believe it could legally throw any of these depending on in which order `starknet_call` performs its internal operations.

starknet_devnet/blueprints/rpc/structures/responses.py
FabijanC: Actually this could be the case if the tx is a part of the currently pending block (it doesn't yet have a block hash).

test/rpc/test_rpc_schema.py
FabijanC: Ok, but what kind of type of error will be raised? Will it result in an internal server error? If not, why can't we assert that actual error?

test/rpc/test_rpc_schema.py
FabijanC: Why not do assertion on entries in `RpcErrorCode` instead of hardcoding ints?

test/rpc/test_rpc_schema.py
cptartur: It is supposed to raise one of the errors documented in rpc for `starknet_call` (assuming we are doing error handling correctly). I can assert it's one of these errors like I did in the method above if you think it's useful.

What is most important in this tests is that no `RpcErrorCode`s are raised - input validated correctly and `call` method actually run.

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,335 @@
...
+import pytest
+from starkware.starknet.public.abi import get_selector_from_name
+
+from starknet_devnet.blueprints.rpc.schema import _assert_valid_rpc_request
+from starknet_devnet.blueprints.rpc.structures.types import PredefinedRpcErrorCode
+from starknet_devnet.blueprints.rpc.utils import rpc_felt
+
+
+def _assert_call_not_raises_predefined_error(error_code: int):
+    assert error_code != PredefinedRpcErrorCode.INVALID_PARAMS.value
```
FabijanC, line 17: Why this negative check?

FabijanC (changes requested):

test/rpc/test_rpc_schema.py
FabijanC: So why did `is_required` have to be introduced if there is `required`?

starknet_devnet/blueprints/rpc/structures/types.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, code, message):
         self.message = message
 
 
-class RpcErrorCode(Enum):
+class PredefinedRpcErrorCode(Enum):
```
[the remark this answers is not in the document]
cptartur: As generally present in the JSON-RPC protocol. I've taken the wording from [RPC specification](https://www.jsonrpc.org/specification) section 5.1, where it says
```
The error codes from and including -32768 to -32000 are reserved for pre-defined errors.
```

test/rpc/test_rpc_schema.py
```diff
@@ -0,0 +1,335 @@
...
+
+import pytest
+from starkware.starknet.public.abi import get_selector_from_name
+
+from starknet_devnet.blueprints.rpc.schema import _assert_valid_rpc_request
+from starknet_devnet.blueprints.rpc.structures.types import PredefinedRpcErrorCode
+from starknet_devnet.blueprints.rpc.utils import rpc_felt
+
+
+def _assert_call_not_raises_predefined_error(error_code: int):
```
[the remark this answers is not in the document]
cptartur: Can change it to `_assert_call_does_not_raise_predefined_error`, maybe its a bit better? 😅 But both are verbose and not very nice

test/rpc/test_rpc_schema.py
cptartur: In the `_request_schemas_for_method`, when I "repackage" the json for validation purposes there's a code like
```py
    for param in params_json:
        name = param["name"]
        is_required = param.get("required", False)
        schema = {
            **param["schema"],
            "is_required": is_required,
            "components": {"schemas": schemas},
        }
        request_schemas[name] = schema
```

I can change it to just `required` from `is_required`. Honestly don't know why I chose to rename it. Can't directly use it though, because I'm not sure if for optional parameters json rpc would have `"required": false` or I'd just be omitted entirely.

test/rpc/test_rpc_schema.py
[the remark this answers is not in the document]
FabijanC: It's ok like this

starknet_devnet/blueprints/rpc/structures/types.py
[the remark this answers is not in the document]
FabijanC: It's ok as it is.

test/rpc/test_rpc_schema.py
cptartur: Oops, it was meant to check agains all possible values. Will fix that.

test/rpc/test_rpc_schema.py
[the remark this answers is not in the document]
cptartur: Changed it anyways 😛

test/rpc/test_rpc_schema.py
FabijanC: I guess it's ok as it is.

cptartur:
I did some basic benchmarking by running `time poetry run pytest -n auto test/rpc` a couple of times agains this PR and master branch. There seems to be a performance penalty (less than 10 seconds for `real` time around 1m15s) for the validation. 
I don't know how this scales with larger tests suites i.e. if the performance hit comes from validation itself or schema creation (which is cached for methods). 

Personally it is not problematic (and validation can be disabled), but that's something to have in mind.

FabijanC (approved):

PR #409, opened 2023-02-24
Improve validation for get_block_by_hash

mikiw:
(cherry picked from commit ad1cd27051d064377c61863b2a53c0c1581293e9)

## Usage related changes

- Change 1.
- ...

- Change 1.
- ...

## Checklist:

- [ ] Applied formatting - `./scripts/format.sh`
- [ ] No linter errors - `./scripts/lint.sh`
- [ ] Performed code self-review
- [ ] Rebased to the last commit of the target branch (or merged it into my branch)
- [ ] Documented the changes
- [ ] Updated the tests

### DIFF
--- a/starknet_devnet/blocks.py
+++ b/starknet_devnet/blocks.py
@@ -32,7 +32,13 @@ def _parse_block_hash(raw: Optional[str]):
         return raw
 
     try:
-        return int(raw, 16)
+        if raw.startswith("0x"):
+            return int(raw, 16)
+
+        raise StarknetDevnetException(
+            code=StarkErrorCode.MALFORMED_REQUEST,
+            message=f"Block hash should be a hexadecimal string starting with 0x, or 'null'; got: '{raw}'.",
+        )
     except ValueError as error:
         raise StarknetDevnetException(
             code=StarkErrorCode.MALFORMED_REQUEST,
--- a/test/test_endpoints.py
+++ b/test/test_endpoints.py
@@ -154,6 +154,12 @@ def get_block_by_number(req_dict: dict):
     )
 
 
+def get_block_by_hash(req_dict: dict):
+    """Get block by block hash"""
+    block_hash = req_dict["blockHash"]
+    return requests.get(f"{APP_URL}/feeder_gateway/get_block?blockHash={block_hash}")
+
+
 def get_transaction_trace(transaction_hash: str):
     """Get transaction trace from request dict"""
     return requests.get(
@@ -232,6 +238,17 @@ def test_error_response_call_with_negative_block_number():
     json_error_message = resp.json()["message"]
     assert resp.status_code == 500
     assert json_error_message is not None
+
+
+@pytest.mark.call
+@devnet_in_background()
+def test_error_response_call_with_block_hash_0():
+    """Call with block hash 0 without 0x prefix should fail"""
+    resp = get_block_by_hash({"blockHash": 0})
+
+    json_error_message = resp.json()["message"]
+    assert resp.status_code == 500
+    assert json_error_message.startswith("Block hash should be a hexadecimal string starting with 0x, or 'null';")
 
 
 @pytest.mark.call


starknet_devnet/blocks.py
```diff
@@ -32,7 +32,13 @@ def _parse_block_hash(raw: Optional[str]):
         return raw
 
     try:
-        return int(raw, 16)
+        if raw.startswith("0x"):
+            return int(raw, 16)
+
+        raise StarknetDevnetException(
+            code=StarkErrorCode.MALFORMED_REQUEST,
+            message=f"Block hash should be a hexadecimal string starting with 0x, or 'null'; got: '{raw}'.",
```
FabijanC, line 40: Is this the message sent by the official feeder gateway? And is `MALFORMED_REQUEST` what's being sent by the gateway?

starknet_devnet/transactions.py
```diff
@@ -194,8 +195,13 @@ def __get_transaction_by_hash(self, tx_hash: str) -> DevnetTransaction or None:
         """
         Get a transaction by hash.
         """
-        numeric_hash = int(tx_hash, 16)
-        return self.__instances.get(numeric_hash)
+        if tx_hash.startswith("0x"):
```
FabijanC, line 198: String starting with `0x` doesn't necessarily make it a hex string. It should be wrapped in a try-except block to prevent Internal Server Errors caused by strings like `0xyz`. E.g.:

```python
if tx_hash.startswith("0x"):
    try:
        return int(tx_id, 16)
    except ValueError:
        pass

raise ...
```

starknet_devnet/blueprints/base.py
```diff
@@ -48,7 +48,7 @@ def extract_positive(request_json, prop_name: str):
     return value
 
 
-def hex_converter(request_json, prop_name: str, convert=custom_int) -> int:
+def hex_converter(request_json, prop_name: str, convert=validate_hex_string) -> int:
```
FabijanC, line 51: I wouldn't expect a function named `validate_...` to return something

starknet_devnet/util.py
```diff
@@ -15,16 +15,20 @@
...
-    Converts the argument to an integer.
-    Conversion base is 16 if `arg` starts with `0x`, otherwise `10`.
+    Converts the argument to an integer only if it starts with `0x`.
     """
-    base = 16 if arg.startswith("0x") else 10
-    return int(arg, base)
+    if arg.startswith("0x"):
+        return int(arg, 16)
+
+    raise StarknetDevnetException(
```
FabijanC, line 28: I see this error being defined in multiple places. Why not just derive a new class from `StarknetDevnetException` with the code and message set in advance?

test/test_l1_l2_mock_messaging.py
```diff
@@ -103,7 +103,7 @@ def test_send_message_to_l2_execute_without_data():
         }
     )
 
-    assert response.status_code == 400
+    assert response.status_code == 500
```
FabijanC, line 106: gateway returns 500?

test/test_endpoints.py
```diff
@@ -352,3 +392,39 @@ def test_get_transaction_trace_of_rejected():
...
     assert resp_body["code"] == str(StarknetErrorCode.NO_TRACE)
     assert resp.status_code == 500
+
+
+@devnet_in_background()
+def test_get_transaction_with_tx_hash_0():
+    """Should fail on get_transaction with hash 0 without 0x prefix"""
+    resp = get_transaction("0")
+    resp_body = resp.json()
+    assert resp_body["message"].startswith(INVALID_TRANSACTION_HASH_MESSAGE)
```
FabijanC, line 402: Why `startswith` and not equality?

test/test_endpoints.py
```diff
@@ -154,13 +157,26 @@ def get_block_by_number(req_dict: dict):
     )
 
 
+def get_block_by_hash(req_dict: dict):
```
FabijanC, line 160: Why should this function accepts a dict instead of a hash?

test/test_endpoints.py
```diff
@@ -352,3 +392,39 @@ def test_get_transaction_trace_of_rejected():
     resp_body = resp.json()
     assert resp_body["code"] == str(StarknetErrorCode.NO_TRACE)
     assert resp.status_code == 500
+
+
+@devnet_in_background()
```
FabijanC, line 397: These tests expecting failure can probably be run without devnet_in_background, just relying on `app.test_client()` (see tests in this file for reference)

FabijanC (changes requested):

starknet_devnet/transactions.py
mikiw: right, I'll update tests and make a fix

starknet_devnet/blueprints/base.py
mikiw: I'll rename this! I missed that during my own review. I just copied the method name from the task and I forgot to change that, sorry.

test/test_l1_l2_mock_messaging.py
mikiw: Is that a problem? Let's discuss this on the call.

test/test_endpoints.py
mikiw: there is ` got: '{arg}'` after message so I used `startswith`

starknet_devnet/util.py
mikiw: Let's discuss it on the call

test/test_endpoints.py
mikiw: Fixed in 353fa8954ff6b4dc95f5d23657f65882acfefcd0

starknet_devnet/blueprints/base.py
mikiw: renamed from 'validate_hex_string' to 'parse_hex_string' in 1dd85e8a5f0ed28166d69cf8ef144d7ecc4df94c

test/test_endpoints.py
mikiw: I had trouble with that let's discuss it on a call

starknet_devnet/transactions.py
mikiw: Fixed in 12a619b372e4cb10a44fa01ad3566430ea5ba5b5

test/test_endpoints.py
mikiw: Fixed in multiply commits just check test_endpoints.py file again

FabijanC (approved):

PR #410, opened 2023-02-25
Make number of retries in forking configurable

### LINKED ISSUE #396: Make retries in forking mode configurable
https://github.com/Shard-Labs/starknet-devnet/blob/94f8d5153b8851283eac3716e282d423875a3c59/starknet_devnet/devnet_config.py#L145

Supporting configurable retrying in forking mode would probably be as simple as replacing this hardcoded `n_retries=1` with something defined by the user on startup.

The user could specify this using `--fork-retry <NUMBER>` with the default value of 1.

bigherc18:
## Usage related changes

- The user could specify the number of retries using `--fork-retry NUMBER` with the default value of 1.
- Close #396 

None

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/starknet_devnet/devnet_config.py
+++ b/starknet_devnet/devnet_config.py
@@ -186,19 +186,21 @@ class NonNegativeAction(argparse.Action):
         setattr(namespace, self.dest, value)
 
 
-class ForkRetriesAction(argparse.Action):
+class PositiveAction(argparse.Action):
     """
-    Action for parsing the --fork-retries int argument, it should be -1, 0 or positive int.
+    Action for parsing positive int argument;
     """
 
     def __call__(self, parser, namespace, values, option_string=None):
-        error_msg = f"argument {option_string} must be either a positive int or equals to -1; got: {values}."
+        error_msg = (
+            f"argument {option_string} must be a positive integer; got: {values}."
+        )
         try:
             value = int(values)
         except ValueError:
             parser.error(error_msg)
 
-        if not (value > 0 or value == -1):
+        if value <= 0:
             parser.error(error_msg)
 
         setattr(namespace, self.dest, value)
@@ -322,8 +324,8 @@ def parse_args(raw_args: List[str]):
         "--fork-retries",
         type=int,
         default=1,
-        action=ForkRetriesAction,
-        help="Specify the number of retries of failed HTTP requests sent to the network before giving up, set to -1 for unlimited retries, defaults to 1",
+        action=PositiveAction,
+        help="Specify the number of retries of failed HTTP requests sent to the network before giving up, defaults to 1",
     )
     parser.add_argument(
         "--chain-id",
--- a/starknet_devnet/devnet_config.py
+++ b/starknet_devnet/devnet_config.py
@@ -359,6 +359,7 @@ class DevnetConfig:
         self.hide_predeployed_accounts = self.args.hide_predeployed_accounts
         self.fork_network = self.args.fork_network
         self.fork_block = self.args.fork_block
+        self.fork_retries = self.args.fork_retries
         self.chain_id = self.args.chain_id
         self.validate_rpc_requests = not self.args.disable_rpc_request_validation
         self.validate_rpc_responses = not self.args.disable_rpc_response_validation
--- a/starknet_devnet/devnet_config.py
+++ b/starknet_devnet/devnet_config.py
@@ -304,7 +304,7 @@ def parse_args(raw_args: List[str]):
         "--fork-retry",
         type=int,
         default=1,
-        help="Specify the number of retries when forking --fork-network; defaults to 1",
+        help="Specify the number of retries of failed HTTP requests sent to the network before giving up, set to -1 for unlimited retries, defaults to 1",
     )
     parser.add_argument(
         "--chain-id",
--- a/starknet_devnet/devnet_config.py
+++ b/starknet_devnet/devnet_config.py
@@ -186,6 +186,24 @@ class NonNegativeAction(argparse.Action):
         setattr(namespace, self.dest, value)
 
 
+class ForkRetriesAction(argparse.Action):
+    """
+    Action for parsing the --fork-retries int argument, it should be -1, 0 or positive int.
+    """
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        error_msg = f"argument {option_string} must be either a positive int or equals to -1; got: {values}."
+        try:
+            value = int(values)
+        except ValueError:
+            parser.error(error_msg)
+
+        if not (value > 0 or value == -1):
+            parser.error(error_msg)
+
+        setattr(namespace, self.dest, value)
+
+
 def parse_args(raw_args: List[str]):
     """
     Parses CLI arguments.
@@ -301,9 +319,10 @@ def parse_args(raw_args: List[str]):
         help="Specify the block number where the --fork-network is forked; defaults to latest",
     )
     parser.add_argument(
-        "--fork-retry",
+        "--fork-retries",
         type=int,
         default=1,
+        action=ForkRetriesAction,
         help="Specify the number of retries of failed HTTP requests sent to the network before giving up, set to -1 for unlimited retries, defaults to 1",
     )
     parser.add_argument(
@@ -333,7 +352,7 @@ def parse_args(raw_args: List[str]):
     if parsed_args.fork_network:
         parsed_args.fork_block = parsed_args.fork_block or "latest"
         parsed_args.fork_network, parsed_args.fork_block = _get_feeder_gateway_client(
-            parsed_args.fork_network, parsed_args.fork_block, parsed_args.fork_retry
+            parsed_args.fork_network, parsed_args.fork_block, parsed_args.fork_retries
         )
 
     return parsed_args
@@ -359,7 +378,6 @@ class DevnetConfig:
         self.hide_predeployed_accounts = self.args.hide_predeployed_accounts
         self.fork_network = self.args.fork_network
         self.fork_block = self.args.fork_block
-        self.fork_retries = self.args.fork_retries
         self.chain_id = self.args.chain_id
         self.validate_rpc_requests = not self.args.disable_rpc_request_validation
         self.validate_rpc_responses = not self.args.disable_rpc_response_validation
--- a/starknet_devnet/devnet_config.py
+++ b/starknet_devnet/devnet_config.py
@@ -137,12 +137,12 @@ def _parse_account_class(class_path: str) -> ContractClassWrapper:
     return ContractClassWrapper(contract_class, class_hash_bytes)
 
 
-def _get_feeder_gateway_client(url: str, block_id: str):
+def _get_feeder_gateway_client(url: str, block_id: str, n_retries: int = 1):
     """Construct a feeder gateway client at url and block"""
 
     feeder_gateway_client = FeederGatewayClient(
         url=url,
-        retry_config=RetryConfig(n_retries=1),
+        retry_config=RetryConfig(n_retries=n_retries),
     )
 
     try:
@@ -301,6 +301,12 @@ def parse_args(raw_args: List[str]):
         help="Specify the block number where the --fork-network is forked; defaults to latest",
     )
     parser.add_argument(
+        "--fork-retry",
+        type=int,
+        default=1,
+        help="Specify the number of retries when forking --fork-network; defaults to 1",
+    )
+    parser.add_argument(
         "--chain-id",
         type=_chain_id,
         default=StarknetChainId.TESTNET,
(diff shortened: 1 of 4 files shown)


bigherc18:
What tests should I write for this ?

bigherc18:
@FabijanC can you please take a look here ?

starknet_devnet/devnet_config.py
```diff
@@ -353,6 +359,7 @@ def __init__(self, args: argparse.Namespace = None):
         self.hide_predeployed_accounts = self.args.hide_predeployed_accounts
         self.fork_network = self.args.fork_network
         self.fork_block = self.args.fork_block
+        self.fork_retries = self.args.fork_retries
```
FabijanC, line 362: Probably no need to introduce this here since it's not used anyhwere else in the code other than `devnet_config.py`. And the property name is wrong: `fork_retries` vs `fork_retry`

FabijanC (changes requested):
> What tests should I write for this ?

Regarding tests, you can take a look at `test/test_fork_cli_params.py` and how good and bad input is tested. You should introduce valid input cases (values like `-1` and `1`) and invalid input cases (values like `foo`, `-2` and `0`). Failing should be graceful.

FabijanC:
@bigherc18

After some more thinking:
- If you agree that it's better, you can use the name `--fork-retries` instead of `--fork-retry`.
- Should `0` be accepted or not? If n_retries means how many times the connection is attempted, then it shouldn't be accepted, but if it's really the number of retries, then `0` should be legal.

starknet_devnet/devnet_config.py
bigherc18: OK. I'll remove it

bigherc18:
- It's already named `--fork-retries`
- I think `0` should be accepted and it means no retry rather than no attempt at all

bigherc18:
After writing tests for the new `--fork-retries` parameter, I encountered a bizarre behavior, during `ìnitialize` we try to deploy the predeployed accounts  to the fork network and fail ! Is this expected ? Am I missing something here ?

```
❯ starknet-devnet --accounts 1 --fork-network alpha-goerli --fork-retries -1
DEBUG:asyncio:Using selector: KqueueSelector
New: FeederGatewayClient(url='https://alpha4.starknet.io', ssl_context=None, retry_config=RetryConfig(n_retries=-1, retry_codes=(<HTTPStatus.BAD_GATEWAY: 502>, <HTTPStatus.SERVICE_UNAVAILABLE: 503>, <HTTPStatus.GATEWAY_TIMEOUT: 504>))) 770261
New: number of retries is set to -1
Account #0
Address: 0x2a3b55a560bc2054c72869e891bc1ae666f18e7cee6433a61f6713ca7ef4ffb
Public key: 0x1eb19309e1a16006a720316e047a7c4dcef05a389b27a588b6fa38650503bf
Private key: 0x9e4980e9cc59a9cd1aa440e67b00d0fc

Initial balance of each account: 1000000000000000000000 WEI
Seed to replicate this account sequence: 1213427098
WARNING: Use these accounts and their keys ONLY for local testing. DO NOT use them on mainnet or other live networks because you will LOSE FUNDS.

DEBUG:asyncio:Using selector: KqueueSelector
New: Forking https://alpha4.starknet.io from block 770261
INFO:starkware.starkware_utils.config_base:Initialized starknet_os_config configuration with value: StarknetOsConfig(chain_id=<StarknetChainId.TESTNET: 1536727068981429685321>, fee_token_address=2087021424722619777119509474943472645767659996348769578120564519014510906823)
INFO:starkware.starkware_utils.config_base:Initialized contract_storage_commitment_tree_height configuration with value: 251
INFO:starkware.starkware_utils.config_base:Initialized global_state_commitment_tree_height configuration with value: 251
INFO:starkware.starkware_utils.config_base:Initialized invoke_tx_max_n_steps configuration with value: 1000000
INFO:starkware.starkware_utils.config_base:Initialized validate_max_n_steps configuration with value: 1000000
INFO:starkware.starkware_utils.config_base:Initialized min_gas_price configuration with value: 100000000000
INFO:starkware.starkware_utils.config_base:Initialized sequencer_address configuration with value: 1556746869602934183907209817909870193193832701883543897870108478262537288705
INFO:starkware.starkware_utils.config_base:Initialized tx_commitment_tree_height configuration with value: 64
INFO:starkware.starkware_utils.config_base:Initialized tx_version configuration with value: 1
INFO:starkware.starkware_utils.config_base:Initialized event_commitment_tree_height configuration with value: 64
INFO:starkware.starkware_utils.config_base:Initialized cairo_resource_fee_weights configuration with value: {'n_steps': 0.05, 'output_builtin': 0.0, 'pedersen_builtin': 1.6, 'range_check_builtin': 0.8, 'ecdsa_builtin': 102.4, 'bitwise_builtin': 3.2, 'ec_op_builtin': 51.2}
New: Returned <starkware.starknet.testing.starknet.Starknet object at 0x10a2e3ee0>
New: deployed fee token
New: account: start deploying account 0x2a3b55a560bc2054c72869e891bc1ae666f18e7cee6433a61f6713ca7ef4ffb
New: account: starknet=<starkware.starknet.testing.starknet.Starknet object at 0x10a2e3ee0>
New: account: set_contract_class
ERROR:services.external_api.client:Got BadRequest while trying to access https://alpha4.starknet.io/feeder_gateway/get_class_hash_at?contractAddress=0x2a3b55a560bc2054c72869e891bc1ae666f18e7cee6433a61f6713ca7ef4ffb&blockNumber=770261., retrying...
Traceback (most recent call last):
  File "/Users/mac/Workspace/OpenSource/starknet-devnet/venv/lib/python3.9/site-packages/services/external_api/client.py", line 117, in _send_request
    raise BadRequest(status_code=response.status, text=text)
services.external_api.client.BadRequest: HTTP error ocurred. Status: 500. Text: {"code": "StarknetErrorCode.UNINITIALIZED_CONTRACT", "message": "Contract with address 0x2a3b55a560bc2054c72869e891bc1ae666f18e7cee6433a61f6713ca7ef4ffb is not deployed."}
^CTraceback (most recent call last):
  File "/Users/mac/Workspace/OpenSource/starknet-devnet/venv/bin/starknet-devnet", line 6, in <module>
    sys.exit(main())
  File "/Users/mac/Workspace/OpenSource/starknet-devnet/starknet_devnet/server.py", line 101, in main
    asyncio.run(state.starknet_wrapper.initialize())
  File "/opt/homebrew/Cellar/python@3.9/3.9.14/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/runners.py", line 44, in run
    return loop.run_until_complete(main)
  File "/opt/homebrew/Cellar/python@3.9/3.9.14/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 634, in run_until_complete
    self.run_forever()
  File "/opt/homebrew/Cellar/python@3.9/3.9.14/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 601, in run_forever
    self._run_once()
  File "/opt/homebrew/Cellar/python@3.9/3.9.14/Frameworks/Python.framework/Versions/3.9/lib/python3.9/asyncio/base_events.py", line 1869, in _run_once
    event_list = self._selector.select(timeout)
  File "/opt/homebrew/Cellar/python@3.9/3.9.14/Frameworks/Python.framework/Versions/3.9/lib/python3.9/selectors.py", line 562, in select
    kev_list = self._selector.control(None, max_ev, timeout)
KeyboardInterrupt
```

bigherc18:
> * OK. I'll use `--fork-retries`
> 
>     * I think `0` should be accepted and it means no retry rather than no attempt at all

`0`is not even accepted by `services.external_api.client.RetryConfig` so I preferred to not accept it

FabijanC:
@bigherc18 I don't understand; does it crash? Does it behave the same on `master`?

bigherc18:
@FabijanC Yes, it behaves exactly the same on `master`, just set the `n_retries` to `-1` and enable debug logs in `main`
```python
import logging
logging.basicConfig(level=logging.DEBUG)
```

It doesn't crash, it hangs forever at [`starknet_devnet/account.py#L60`](https://github.com/bigherc18/starknet-devnet/blob/b6947da70747a1eb91bcf18178e90f69068bf274/starknet_devnet/account.py#L60)

Running:
```
starknet-devnet --fork-network alpha-goerli2
```

will keep logging this error forever:

```
ERROR:services.external_api.client:Got BadRequest while trying to access https://alpha4-2.starknet.io/feeder_gateway/get_class_hash_at?contractAddress=0x74872cea566e072c70cbfe536f70cb362652270637936def6101660a560b872&blockNumber=89624., retrying...
```

FabijanC:
@bigherc18 and if you set it to e.g. 5, it will do this 5 times?

bigherc18:
> @bigherc18 and if you set it to e.g. 5, it will do this 5 times?

Yes

FabijanC:
@bigherc18 Thank you for discovering and sharing this issue, but I cannot pay full attention to it now. Perhaps somehow this predeployment would need to be modified in case of forking, but I don't have an exact good idea.

 If you can try doing more research, we would be grateful, otherwise I will do it once I have the time.

bigherc18:
@FabijanC I'll investigate this more and get back to you. But, since it's not related to the current PR, I suggest we merge this and open a separate issue/PR ?

FabijanC:
You can open a new issue, but we can simply keep this unmerged.

bigherc18:
I'm investigating the issue and having difficulties to understand how things works here.

Can you please explain to me how forking is implemented ?

FabijanC:
> Can you please explain to me how forking is implemented ?

When Devnet doesn't find something in its CachedState, it fetches from the forked network (using the FeederGatewayClient in forked_state.py). Upon fetching, it stores the retrieved piece of data for further use.

To achieve this, we didn't have to modify the way things are written into Starknet's state, only the way they are read (that's why the StateReader was inherited).

bigherc18:
After investigation, here's what I got:

> When Devnet doesn't find something in its CachedState, it fetches from the forked network

Unfortunately, this also applies for predeployed accounts, specifically when calling:

`Account.deploy` -> `CachedState.deploy_contract` -> `CachedState.get_class_hash_at` -> `ForkedStateReader.get_class_hash_at` -> `FeederGatewayClient.get_class_hash_at`

Obviously, the predeployed contract account is not available in the forked network and the last call will fail with a `BadRequest` which is caught by `ForkedState` and returns `UNINITIALIZED_CLASS_HASH`, But when we set `--fork-retries=-1` the `FeederGatewayClient.get_class_hash_at` will hang indefinitely trying to get the class hash of the non existent predeployed account contract.

Sending request to `/feeder_gateway/get_class_hash_at` for a non existent contract address will cause the same problem of `FeederGatewayClient.get_class_hash_at` hanging indefinitely trying to fetch it from the forked network

Also, the same thing happen when calling `StarknetWrapper.is_deployed` for chargeable account because it calls `ForkedStateReader.get_class_hash_at` (Side question: why do we need to call `is_deployed` only for chargeable account but not for accounts and fee token ?)

So, what we should do is first to not accept `--fork-retries=-1`

This problem doesn't happen for fee token because fee token doesn't call `deploy_account`, instead it caches the class hash directly in `starknet.state.state.cache._class_hash_writes`

What we can do is either ignore this for now as it won't do any harm as long as we don't accept `--fork-retries=-1` or avoid calling `ForkedStateReader.get_class_hash_at` on predeployed accounts and do the same as we're doing in fee token

@FabijanC What do you think ?

FabijanC:
@bigherc18 Thanks for investigating!

If you replace `state.deploy_contract` with direct `_class_hash_writes` manipulation, do all the tests still pass then (`./scripts/test.sh`)?

Also, I wouldn't say that just not accepting `--fork-retries=-1` removes any harm, e.g. if the value of 10 is provided to `--fork-retries`, doesn't that slow down (by how much?) Devnet's startup?

bigherc18:
@FabijanC Yes, all the tests pass if we avoid `state.deploy_contract` 

Note that if we want to not hit the forked network at all during startup we'll also need to avoid/rewrite `is_deployed` call for `__deploy_chargeable_account` as it also calls `get_class_hash_at` for a predeployed account which doesn't exist in the forked network

bigherc18:
No it doesn't slow down the startup, even if we pass `100000000` it only adds ~0.5s to the startup time (measured with `time` command`) because no matter what the value is, as long as it's not `-1`, `ClientBase._send_request` function defined in `services/external_api/client.py` will not retry `BadRequest` errors but if the value is `-1`, it'll keep retrying indefinitely, here's the if clause it uses to decide whether to retry or not
```python
if limited_retries and (
    n_retries_left == 0
     or exception.status_code not in self.retry_config.retry_codes
):
``` 
With:
```python
limited_retries = self.retry_config.n_retries > 0
# n_retries > 0 means limited retries; n_retries == -1 means unlimited retries.
```

FabijanC:
> No it doesn't slow down the startup, even if we pass 100000000 it only adds ~0.5s to the startup time 

In that case, I'm ok with just disallowing -1

bigherc18:
> In that case, I'm ok with just disallowing -1

Do you want to just do that and don't manipulate `_class_hash_writes` instead of calling `deploy_contract` ?

FabijanC:
> > In that case, I'm ok with just disallowing -1
> 
> Do you want to just do that and don't manipulate `_class_hash_writes` instead of calling `deploy_contract` ?

If you don't mind, yes

FabijanC (approved):

PR #414, opened 2023-02-28
Refactor predeployed contract wrapper classes

### LINKED ISSUE #325: Refactor predeployed contract wrapper classes
UDC, FeeToken, Account (+ChargeableAccount) - these classes have common logic (namely the `deploy` method). They could use a common parent class.

bigherc18:
## Usage related changes

No change affects the users

Predeployed contract wrapper classes now have a common parent class called `PredeployedContractWrapper` that should be inherited if we add other classes in the future.

Closes #325

## Checklist:

- [ ] Documented the changes
- [ ] Updated the tests


starknet_devnet/predeployed_contract_wrapper.py
```diff
@@ -0,0 +1,9 @@
+"""Parent for predeployed contract wrapper classes"""
+
+
+class PredeployedContractWrapper:
+    """Parent class for all predeployed contract wrapper classes"""
+
+    async def deploy(self):
```
FabijanC, line 7: I would expect this base class to also do some common logic, otherwise I don't see a lot of use from this abstraction. I.e. this could be a template method (which does e.g. starknet state manipulation) and there could be some unimplemented hook methods (e.g. one could do constructor mocking).

FabijanC (changes requested):

starknet_devnet/predeployed_contract_wrapper.py
bigherc18: What do you think of something like this:
```python
class PredeployedContractWrapper:
    """Parent class for all predeployed contract wrapper classes"""

    async def deploy_contract(self):
        # pylint: disable=protected-access
        starknet: Starknet = self.starknet_wrapper.starknet
        starknet.state.state.cache._class_hash_writes[
            self.address
        ] = self.class_hash_bytes
        # replace with await starknet.state.state.deploy_contract

    async def mimic_constructor(self):
        raise NotImplementedError()

    async def deploy(self):
        """Deploy the contract wrapper to devnet"""
        starknet: Starknet = self.starknet_wrapper.starknet

        await starknet.state.state.set_contract_class(
            self.class_hash_bytes, self.contract_class
        )

        await self.deploy_contract()

        await self.mimic_constructor()

        self.contract = StarknetContract(
            state=starknet.state,
            abi=self.contract_class.abi,
            contract_address=self.address,
            deploy_call_info=None,
        )
```

starknet_devnet/predeployed_contract_wrapper.py
```diff
@@ -1,9 +1,100 @@
...
+        except AttributeError:
+            try:
+                return self.HASH_BYTES
+            except AttributeError:
+                raise NotImplementedError(
+                    "Predeployed contracts should provide a class_hash_bytes attribute or a HASH_BYTES class attribute"
+                )
+
+    @class_hash_bytes.setter
+    def class_hash_bytes(self, class_hash_bytes: bytes):
```
FabijanC, line 46: I don't understand the need for this? Is this approach a Python idiom?

starknet_devnet/predeployed_contract_wrapper.py
```diff
@@ -1,9 +1,100 @@
...
+            except AttributeError:
+                raise NotImplementedError(
+                    "Predeployed contracts should provide a contract_class attribute or get_contract_class method"
+                )
+
+    @contract_class.setter
+    def contract_class(self, contract_class):
+        self._contract_class = contract_class
+
+    async def deploy_contract(self):
```
FabijanC, line 65: I don't like it that we have both `deploy_contract` and `deploy`. If some of these methods are supposed to be protected (only accessed within the class and its children), maybe they can be prefixed with an underscore (`_`).

starknet_devnet/predeployed_contract_wrapper.py
```diff
@@ -1,9 +1,100 @@
...
+        # pylint: disable=protected-access
+        starknet: Starknet = self.starknet_wrapper.starknet
+        starknet.state.state.cache._class_hash_writes[
+            self.address
+        ] = self.class_hash_bytes
+        # replace with await starknet.state.state.deploy_contract
+        # await starknet.state.state.deploy_contract(self.address, self.class_hash_bytes)
+        # For now, it fails for fee token since the address is the same as the
```
FabijanC, line 73: Fee Token is Devnet's ETH Token, it's supposed to be on the same address

starknet_devnet/predeployed_contract_wrapper.py
```diff
@@ -1,9 +1,100 @@
...
+from starkware.starknet.testing.starknet import Starknet
+from starkware.starknet.services.api.contract_class import ContractClass
+from starkware.starknet.testing.contract import StarknetContract
 
-class PredeployedContractWrapper:
+
+class PredeployedContractWrapper(ABC):
     """Parent class for all predeployed contract wrapper classes"""
 
+    # Cannot import it because of circular imports
```
FabijanC, line 12: Regarding this, you can see how it has been handled so far in e.g. account.py

FabijanC (changes requested):

starknet_devnet/predeployed_contract_wrapper.py
bigherc18: We need the setter to be able to set the value inside `__init__` with `self.class_hash_bytes = ...` because `class_hash_bytes` is a `property`

This way, classes can either provide a `HASH_BYTES` or set the value directly inside `__main__` or even override the `property` altogether

starknet_devnet/predeployed_contract_wrapper.py
bigherc18: We can even get rid of the function, its there so that we can easily switch between calling `state.deploy_contract` and manipulating the cache

BTW, what do we do about it ? Since you prefer calling `state.deploy_contract` but it doesn't work, obviously, for Fee Token ?

starknet_devnet/predeployed_contract_wrapper.py
bigherc18: The comment is there because it's not obvious why we're manipulating the cache and not calling `state.deploy_contract`, I spent some time understanding why and figured it'll be good to have that in the comment

starknet_devnet/predeployed_contract_wrapper.py
FabijanC: Can all predeployed contracts directly manipulate the underlying object instead of calling `deploy_contract`?

Also, sure I prefer calling public methods rather than modifying private properties which shouldn't be modified and which make me suppress the linter xD But if they don't work, they don't work.

bigherc18: Yes.

FabijanC: Ok, then I'd consider all of the predeployed contracts deploying so (without `state.deploy_contract`), especially if it will be done in just one place (which it should since we are introducing this common parent class).

starknet_devnet/predeployed_contract_wrapper.py
bigherc18: All imports doesn't work here because of:

```
ImportError while loading conftest '/Users/mac/Workspace/OpenSource/starknet-devnet/test/conftest.py'.
test/conftest.py:10: in <module>
    from .util import run_devnet_in_background, terminate_and_wait
test/util.py:21: in <module>
    from starknet_devnet.general_config import DEFAULT_GENERAL_CONFIG
starknet_devnet/general_config.py:15: in <module>
    from .fee_token import FeeToken
starknet_devnet/fee_token.py:14: in <module>
    from starknet_devnet.chargeable_account import ChargeableAccount
starknet_devnet/chargeable_account.py:5: in <module>
    from starknet_devnet.account import Account
starknet_devnet/account.py:14: in <module>
    from starknet_devnet.predeployed_contract_wrapper import PredeployedContractWrapper
starknet_devnet/predeployed_contract_wrapper.py:9: in <module>
    class PredeployedContractWrapper(ABC):
starknet_devnet/predeployed_contract_wrapper.py:13: in PredeployedContractWrapper
    from .starknet_wrapper import StarknetWrapper
starknet_devnet/starknet_wrapper.py:60: in <module>
    from .accounts import Accounts
starknet_devnet/accounts.py:11: in <module>
    from .account import Account
E   ImportError: cannot import name 'Account' from partially initialized module 'starknet_devnet.account' (most likely due to a circular import) (/Users/mac/Workspace/OpenSource/starknet-devnet/starknet_devnet/account.py)
```

starknet_devnet/predeployed_contract_wrapper.py
bigherc18: So, what do we do with this ?

FabijanC: I refuse to believe that nested try-except blocks are a good approach. There is a global function called `hasattr`, but I would avoid even using that.

The base class could specify unimplemented (abstract) getters, and child classes could implement them by returning the precalculated values that are stored as their class properties (e.g. `HASH`). I haven't tried coding this proposal, so I don't know how it looks, but it sounds better to me.

FabijanC (changes requested):

FabijanC:
Also, can you update your branch? It is out-of-date with the base branch

starknet_devnet/predeployed_contract_wrapper.py
```diff
@@ -0,0 +1,97 @@
...
+                raise NotImplementedError(
+                    "Predeployed contracts should provide a contract_class attribute or get_contract_class method"
+                )
+
+    @contract_class.setter
+    def contract_class(self, contract_class):
+        self._contract_class = contract_class
+
+    async def mimic_constructor(self):
+        pass
```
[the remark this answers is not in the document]
bigherc18: No. Since not all subclasses needs to override it, ex: UDC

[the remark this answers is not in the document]
FabijanC: I think it should because child classes should be the ones with an empty implementation. The method in the parent class should be abstract (but not by means of `pass`ing).

Also, if `mimic_constructor` is only used inside the class (or its children), it should be prefixed with an underscore (`_mimic_constructor`) to show protectedness.

starknet_devnet/fee_token.py
```diff
@@ -26,10 +27,12 @@ class FeeToken:
     # HASH = to_bytes(compute_class_hash(contract_class=FeeToken.get_contract_class()))
     HASH = 0x6A22BF63C7BC07EFFA39A25DFBD21523D211DB0100A0AFD054D172B81840EAF
     HASH_BYTES = to_bytes(HASH)
+    class_hash_bytes = HASH_BYTES
 
     # Taken from
     # https://github.com/starknet-community-libs/starknet-addresses/blob/df19b17d2c83f11c30e65e2373e8a0c65446f17c/bridged_tokens/goerli.json
     ADDRESS = 0x49D36570D4E46F48E99674BD3FCC84644DDD6B96F7C741B1562B82F9E004DC7
+    address = ADDRESS
```
FabijanC, line 35: Seems redundant to have both `ADDRESS` and `address` (same for hash bytes). Can you make it so that we have just one?

FabijanC (commented):
Ok, looks nicer! Just left one comment.

starknet_devnet/fee_token.py
bigherc18: That's why I did it the way it was before, to avoid redundancy.

Do you suggest we remove `ADDRESS` in favor of `address` and replace every use of it in the code base ?

FabijanC: Can PredeployedAccount parent class can depend on `ADDRESS` ?

bigherc18: That'd make it look like a constant which isn't in the case of `Account`

FabijanC: Ok, good point. In that case there should be an **abstract** getter: either something like `get_address` or a formally annotated property (can the latter even be abstract?). Then all child classes other than `Account` should implement it by returning their respective precalculated constants and `Account` can return `self.address` (or it should be `self._address`).

bigherc18: We can keep the predeployed class as it is and replace
```python
address = ADDRESS 
```
With
```python
@property
def address(self) -> int:
    return self.ADRRESS
```

FabijanC: Looks good!

FabijanC: I think it's ok.

(It is kind of a compromise between the `@property` approach and having `address` and `ADDRESS` properties next to each other, but it's reasonable IMO)

FabijanC (approved):
If you can tell me that `./scripts/test.sh` is passing, we can merge this. If we don't merge this today, we probably will by Tuesday.

Thanks for the cooperation and all the suggestions!

bigherc18:
@FabijanC Yes, tests are passing

Issue #417, opened 2023-03-04
Add a command to abort blocks


fracek:
At the moment, devnet supports the `/create_block` and `/create_block_on_demand` commands to add blocks to the chain. These methods are incredible for testing anything that integrates with the RPC since we can use them to construct any type of block.

It would be incredibly useful to add a `/abort_blocks_after` command to mark all blocks after the specified as aborted. Like in the StarkNet gateway/RPC, clients can access these blocks only by hash but not by number (since they don't belong to the canonical chain anymore).

This feature enables developers to test their code to handle chain reorganizations, which infrequently happen on StarkNet.

mikiw:
@fracek Let's say we have blocks:
- 0x1 (ACCEPTED_ON_L2)
- 0x2 (ACCEPTED_ON_L2)
- 0x3 (ACCEPTED_ON_L2)

so the expectation of calling `/abort_blocks_after?blockHash=0x2` is that state of the blockchain will be updated and after calling `get_block` we will have changed state (also new block can be mined like 0x4):
- 0x1 (ACCEPTED_ON_L2)
- 0x2 (ABORTED)
- 0x3 (ABORTED)
- 0x4 (ACCEPTED_ON_L2)

is that correct?

mikiw:
@fracek what should happen with the status of transactions whose block has been aborted? Transactions should be rejected?

fracek:
We should double check with nodes, because the same transaction can be included in the new blocks if the previous were rejected.

mikiw:
How can we do that? There is no other nodes concept in devnet.

Also during `/abort_blocks_after` call, we don't know if these transactions will be in next block/blocks.

mikiw:
During the development, I encountered this code:
```
class BaseResponseObject:
    """
    Contains common functionality to response objects from the FeederGateway.
    """

    @marshmallow.post_dump
    def remove_none_values(self, data: Dict[Any, Any], many: bool = False) -> Dict[Any, Any]:
        return {key: value for key, value in data.items() if value is not None}
```

which is preventing of `block.dump()` with aborted status which is valid block...

mikiw:
@fracek are you sure that these block should be `aborted` not `rejected`? For RPC in pathfinder node they use only `rejected` status and the have this mapping `Aborted => BlockStatus::Rejected` 

```
    impl From<starknet_gateway_types::reply::Status> for BlockStatus {
[...]

PR #421, opened 2023-03-20
Abort blocks

mikiw:
## Usage related changes

- Add `/abort_blocks_after` endpoint which can abort blocks and transactions from a given block to the latest block

- `abort_block_by_hash()` and `reject_transaction()` functions added
- new tests to the abort block functionality were introduced

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


page/docs/guide/abort-blocks.md
```diff
@@ -0,0 +1,25 @@
...
+sidebar_position: 19
+---
+# Abort blocks
+
+This functionality allows to simulate blocks abort that can occur on mainnet.
+
+
+## Abort Blocks After
+
+Abort blocks and reject transactions from given block hash to last block.
```
FabijanC, line 11: Here's my suggestion, with an example inspired by the example from #417.

```suggestion
Abort blocks and reject transactions from the specified block to the currently latest block. Newly created blocks after the abortion will have accepted status and will continue with numbering where the last accepted block left off.

E.g. assume there are 3 accepted blocks numbered 1, 2 and 3 and your request aborts blocks starting with block 2. This will make blocks 2 and 3 aborted and their transactions rejected. Once a new block is mined, it will be accepted and it will have number 2
```

starknet_devnet/blocks.py
```diff
@@ -306,3 +306,22 @@ async def store_pending(
...
+            block = await self.get_by_number(block_number)
+
+            block_dict = block.dump()
+            block_dict["status"] = BlockStatus.ABORTED.name
+            block_dict["transaction_receipts"] = None
+            self.__num2block[block_number] = StarknetBlock.load(block_dict)
+
+            return block.block_hash
+
+        return ""
```
FabijanC, line 327: What is this?

starknet_devnet/blocks.py
```diff
@@ -306,3 +306,22 @@ async def store_pending(
...
+        """
+        Abort block by given block hash.
+        """
+        numeric_hash = _parse_block_hash(block_hash)
+
+        if numeric_hash in self.__hash2num:
+            block_number = self.__hash2num[int(block_hash, 16)]
+            block = await self.get_by_number(block_number)
+
+            block_dict = block.dump()
```
FabijanC, line 320: Add a comment here that this is done as it is because block object's properties cannot be modified.

starknet_devnet/blueprints/base.py
```diff
@@ -221,3 +221,29 @@ async def fork_status():
     if config.fork_network:
         return jsonify({"url": config.fork_network.url, "block": config.fork_block})
     return jsonify({})
+
+
+@base.route("/abort_blocks_after", methods=["POST"])
+async def abort_blocks_after():
+    """Abort blocks and transactions from given block hash to last block."""
+    request_json = request.json or {}
+    current_block = await state.starknet_wrapper.blocks.get_by_hash(
```
FabijanC, line 230: Shouldn't it be called `starting_block`?

FabijanC: What happens if it's an invalid hash? What if it's a hash of an already aborted block?

FabijanC (changes requested):

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,125 @@
...
+@pytest.mark.usefixtures("run_devnet_in_background")
+@pytest.mark.parametrize(
+    "run_devnet_in_background, expected_block_hash",
+    [
+        (
+            [*PREDEPLOY_ACCOUNT_CLI_ARGS],
+            "",
+        ),
+        (
+            [*PREDEPLOY_ACCOUNT_CLI_ARGS, "--lite-mode"],
```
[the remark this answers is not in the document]
mikiw: Fixed in 8f5239407886ab4496228edffbc03a1aa0711d16

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,125 @@
...
+from .util import (
+    assert_transaction,
+    assert_tx_status,
+    deploy,
+    devnet_in_background,
+    get_block,
+)
+
+EXPECTED_ABORT_HASHES = ["0x1", "0x2"]
+NOT_EXISTING_BLOCK = "0x9"
```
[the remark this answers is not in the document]
mikiw: fixed in 58fb438aa0ae29430e716d8ef94431a8dd2348fd

page/docs/guide/abort-blocks.md
```diff
@@ -0,0 +1,25 @@
...
+
+This functionality allows to simulate blocks abort that can occur on mainnet.
+
+
+## Abort Blocks After
+
+Abort blocks and reject transactions from given block hash to last block. 
+
+```
+POST /abort_blocks_after
```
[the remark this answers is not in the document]
mikiw: Fixed in 627d9e836f2bfd2794f3a3c1f8ed162b55caed75

starknet_devnet/blocks.py
mikiw: fixed in 86a284cbc37ec589309d16f26d570557f634fa7e

starknet_devnet/transactions.py
```diff
@@ -284,6 +284,18 @@ async def get_transaction_status(self, tx_hash: str):
 
         return status_response
 
+    async def reject_transaction(self, tx_hash: int):
+        """
+        Reject transaction in aborted block.
+        """
+        self.__instances[tx_hash].status = TransactionStatus.REJECTED
+        self.__instances[tx_hash].transaction_index = 0
```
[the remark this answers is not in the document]
mikiw: it turned out that this can be removed, fixed in f0d8909dc1e49926a149e8cc46fd7cabf2264505

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,125 @@
...
+    invoke_tx_hash = invoke(
+        calls=[(contract_deploy_info["address"], "increase_balance", [10, 20])],
+        account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
+        private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+    )
+    invoke_block = get_block(parse=True)
+    assert invoke_block["status"] == "ACCEPTED_ON_L2"
+    assert_transaction(invoke_tx_hash, "ACCEPTED_ON_L2")
+
+    # After abort blocks call blocks should be aborted and
```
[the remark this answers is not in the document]
mikiw: Fixed in f0d8909dc1e49926a149e8cc46fd7cabf2264505

starknet_devnet/blocks.py
mikiw: I wanted to return an empty string if condition `if numeric_hash in self.__hash2num` is not met but it can be `None` so I removed it here 9c613510e84280d0ffab641bab272a84c93c05ca

starknet_devnet/blueprints/base.py
mikiw: Fixed in 48eeb54bb5b0f00028e79eb19013862f83a070a2

With invalid hash StarknetErrorCode.BLOCK_NOT_FOUND exception will be raised. 

Wit calling the endpoint twice, the operation will be repeated, and there is no additional check if this is what you asking for.

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,125 @@
...
+    assert invoke_block_after_abort["status"] == "ABORTED"
+    assert_transaction(invoke_tx_hash, "REJECTED")
+
+    # Test RPC get block status
+    rpc_response = rpc_call("starknet_getBlockWithTxs", params={"block_id": "latest"})
+    assert rpc_response["result"]["status"] == "REJECTED"
+
+    # Newly deployed contract after abort should be accepted on L2
+    contract_deploy_info = deploy(CONTRACT_PATH, inputs=["0"])
+    contract_deploy_block = get_block(parse=True)
```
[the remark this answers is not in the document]
mikiw: Still fixing... stay tuned.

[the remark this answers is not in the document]
mikiw: Fixed in the last commits with the change of block data structure, abort logic, and tests.

Please check block.py.

page/docs/guide/abort-blocks.md
```diff
@@ -0,0 +1,27 @@
+---
+sidebar_position: 19
```
FabijanC, line 2: Instead of creating a new documentation file, I think this would make addition to the existing blocks.md file.

starknet_devnet/blocks.py
```diff
@@ -77,22 +77,31 @@ class DevnetBlocks:
...
+        return len(self.__hash2block) + self.origin.get_number_of_blocks()
+
+    def get_number_of_not_aborted_blocks(self) -> int:
+        """Returns the number of not aborted blocks."""
+        not_aborted_blocks = {
+            key: val
+            for key, val in self.__hash2block.items()
+            if val.status is not BlockStatus.ABORTED
+        }
+        return len(dict(not_aborted_blocks)) + self.origin.get_number_of_blocks()
```
FabijanC, line 104: Unnecessary dict wrapper, not_aborted_blocks is already a dict

starknet_devnet/blocks.py
```diff
@@ -77,22 +77,31 @@ class DevnetBlocks:
...
         """Returns the number of blocks stored so far."""
-        return len(self.__num2block) + self.origin.get_number_of_blocks()
+        return len(self.__hash2block) + self.origin.get_number_of_blocks()
+
+    def get_number_of_not_aborted_blocks(self) -> int:
+        """Returns the number of not aborted blocks."""
+        not_aborted_blocks = {
+            key: val
+            for key, val in self.__hash2block.items()
+            if val.status is not BlockStatus.ABORTED
```
FabijanC, line 102: Is it ok to compare enums by identity (i.e. using the `is` operator instead of `==`)?

starknet_devnet/starknet_wrapper.py
```diff
@@ -619,13 +619,12 @@ async def __get_query_state(self, block_id: BlockId = DEFAULT_BLOCK_ID):
         assert isinstance(block_id, dict)
         if block_id.get("block_hash"):
             # takes care of parsing
-            block = await self.blocks.get_by_hash(block_id["block_hash"])
-            return self.blocks.get_state(block.block_number)
+            return self.blocks.get_state(int(block_id["block_hash"], base=16))
```
FabijanC, line 622: The old implementation took care of parsing inside of `get_by_hash` (that's what the comment in L621 says). Now it no longer checks for `block_id["block_hash"]` validity but immediately tries parsing, which might fail.

starknet_devnet/blueprints/rpc/structures/types.py
```diff
@@ -113,5 +113,5 @@ class PredefinedRpcErrorCode(Enum):
 RPC_ERRORS = json.loads(RPC_SPECIFICATION)["components"]["errors"]
 
 GATEWAY_TO_RPC_ERROR = {
-    StarknetErrorCode.OUT_OF_RANGE_BLOCK_ID: RPC_ERRORS["BLOCK_NOT_FOUND"],
```
FabijanC, line 116: Why change this? Shouldn't we just add a new entry?

starknet_devnet/blocks.py
```diff
@@ -77,22 +77,31 @@ class DevnetBlocks:
...
-        number_of_blocks = self.get_number_of_blocks()
+        number_of_blocks = self.get_number_of_not_aborted_blocks()
         return await self.get_by_number(number_of_blocks - 1)
 
     def get_number_of_blocks(self) -> int:
         """Returns the number of blocks stored so far."""
-        return len(self.__num2block) + self.origin.get_number_of_blocks()
+        return len(self.__hash2block) + self.origin.get_number_of_blocks()
+
+    def get_number_of_not_aborted_blocks(self) -> int:
```
FabijanC, line 97: Isn't the number of not aborted blocks (why not call them accepted?) the len of num2hash (+ blocks in origin)?

starknet_devnet/blocks.py
```diff
@@ -103,8 +112,8 @@ def __assert_block_number_in_range(self, block_number: BlockIdentifier):
                 code=StarkErrorCode.MALFORMED_REQUEST, message=message
             )
 
-        if block_number >= self.get_number_of_blocks():
-            message = f"Block number too high. There are currently {len(self.__num2block)} blocks; got: {block_number}."
+        if block_number >= self.get_number_of_not_aborted_blocks():
```
FabijanC, line 115: I don't like it that we are executing get_number_of_not_aborted_blocks multiple times here.

starknet_devnet/blocks.py
```diff
@@ -77,22 +77,31 @@ class DevnetBlocks:
...
         self.__pending_signatures: Sequence[List[int]] = None
         self.__state_archive = MemoryStateArchive()
 
     async def get_last_block(self) -> StarknetBlock:
         """Returns the last block stored so far."""
-        number_of_blocks = self.get_number_of_blocks()
+        number_of_blocks = self.get_number_of_not_aborted_blocks()
         return await self.get_by_number(number_of_blocks - 1)
 
     def get_number_of_blocks(self) -> int:
```
FabijanC, line 93: Why do we need both `get_number_of_blocks` and `get_number_of_not_aborted_blocks`?

I'm worried because rpc/blocks.py is relying on this function and its semantics is being altered.

starknet_devnet/blocks.py
```diff
@@ -168,14 +176,15 @@ async def get_state_update(
...
+            if numeric_hash in self.__state_updates:
+                return self.__state_updates[numeric_hash]
 
-            return await self.origin.get_state_update(block_number=block_number)
+            return await self.origin.get_state_update(block_hash=numeric_hash)
 
         # now we know the block ID is "latest"
         return (
-            self.__state_updates.get(self.get_number_of_blocks() - 1)
+            self.__state_updates[(await self.get_last_block()).block_hash]
```
FabijanC, line 187: The part after `or` gets evaluated if the part before `or` is falsy (e.g. `None`). By using the `[key]` approach, we are risking `KeyError`, and by using the `.get(key)` approach, it really can be `None`.

Although if the comment "now we know the block ID is 'latest'" is true, then I'm not sure when this _defaulting when None_ is useful.

starknet_devnet/blueprints/base.py
```diff
@@ -221,3 +221,37 @@ async def fork_status():
     if config.fork_network:
         return jsonify({"url": config.fork_network.url, "block": config.fork_block})
     return jsonify({})
+
+
+@base.route("/abort_blocks", methods=["POST"])
+async def abort_blocks():
+    """Abort blocks and transactions from given block hash to last block."""
```
FabijanC, line 228: So far we haven't had the practice to put this much logic into a blueprint (blueprints/base.py in this case), usually starknet_wrapper.py, blocks.py and transactions.py were the ones holding the logic. Although this logic does seem correct.

starknet_devnet/blueprints/base.py
```diff
@@ -221,3 +221,37 @@ async def fork_status():
...
+    request_json = request.json or {}
+    starting_block = await state.starknet_wrapper.blocks.get_by_hash(
+        hex(hex_converter(request_json, "startingBlockHash"))
+    )
+    last_block = await state.starknet_wrapper.blocks.get_last_block()
+    blocks_to_abort = []
+    aborted_blocks = []
+
+    # Get blocks to abort
+    for block_number in range(starting_block.block_number, last_block.block_number + 1):
```
FabijanC, line 238: I don't know if this is the optimal way, but if you are going to want to change something, you might want to consider consider relying on `parent_block_hash` property of `StarknetBlock`. That way we could just get the latest block and iterate backwards until the starting block is reached.

FabijanC (changes requested):
What if trying to abort an old block when forking (old in the sense that it was mined on the origin)?

Do we set the latest state after aborting? And drop the pending state (if it exists)?

I have yet to review the testing.

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,143 @@
...
+    ]
+    last_block = get_block(parse=True)
+    assert last_block["status"] == "ACCEPTED_ON_L2"
+    assert last_block["block_number"] == 0
+    contract_deploy_block_after_abort = get_block(
+        block_hash=contract_deploy_block["block_hash"], parse=True
+    )
+    assert contract_deploy_block_after_abort["status"] == "ABORTED"
+    assert_transaction(contract_deploy_info["tx_hash"], "REJECTED")
+    invoke_block_after_abort = get_block(
```
FabijanC, line 115: For better readability, I think it would be useful to put newlines after groups of assertion; i.e. to visually separate deploy-block and invoke-block assertion.

FabijanC (changes requested):
The tests look good, but as noted in the previous review (with checking if test is restored properly after abortion), we could also test if a contract's deployment is reverted and/or the modification (increment) of its storage variables.

starknet_devnet/blueprints/base.py
mikiw: I think that this can be resolved

test/test_abort_blocks_after.py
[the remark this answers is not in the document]
mikiw: I think that this can be resolved

starknet_devnet/blocks.py
mikiw: this `return ""` is outdate

we are testing the unhappy path in `test_abort_not_existing_block()` test

starknet_devnet/blocks.py
mikiw: I take a closer look and actually we just need `get_number_of_accepted_blocks()`

page/docs/guide/abort-blocks.md
```diff
@@ -0,0 +1,27 @@
...
+---
+# Abort blocks
+
+This functionality allows to simulate blocks abort that can occur on mainnet.
+
+
+## Abort Blocks After
```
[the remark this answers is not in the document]
mikiw: I thought that that is a general template but I unified that.

Fixed in 4e241f93a99b8613f712a613abacd8f0d556d88c

page/docs/guide/abort-blocks.md
mikiw: Fixed in a669879e403e242fc80e58776c35390769824833

starknet_devnet/blocks.py
mikiw: Fixed in 4581da40b77aa79ba1f6a3aabd024a63f764feca

starknet_devnet/blocks.py
mikiw: yes it's ok -> https://docs.python.org/3/howto/enum.html#comparisons

I also tested that during programing

starknet_devnet/starknet_wrapper.py
mikiw: @FabijanC so this should not be:
```
return self.blocks.get_state(int(block_id["block_hash"], base=16))
```
but
```
return self.blocks.get_state(int(block_id.get("block_hash"), base=16))
```
?

starknet_devnet/blueprints/rpc/structures/types.py
mikiw: After the last changes, this code was not working as expected on `get` with `ex.code` in `rpc_error_dict`:
```
except StarknetDevnetException as ex:
        default_error = PredefinedRpcErrorCode.INTERNAL_ERROR
        rpc_error_dict = GATEWAY_TO_RPC_ERROR.get(ex.code, default_error)
        return rpc_error(message_id=message_id, **rpc_error_dict)
```
and it's the only one place where `GATEWAY_TO_RPC_ERROR` is used so I decided to change this

starknet_devnet/blocks.py
mikiw: true, fixed in cf60a08603da192c7030c4479e383b187f717a80

starknet_devnet/blocks.py
mikiw: Fixed in cf60a08603da192c7030c4479e383b187f717a80, I updated the comment - it was misleading since we are getting state from block_hash now, not block_number

starknet_devnet/blueprints/base.py
mikiw: I'm changing that

starknet_devnet/transactions.py
```diff
@@ -284,6 +284,17 @@ async def get_transaction_status(self, tx_hash: str):
...
 
+    async def reject_transaction(self, tx_hash: int):
+        """
+        Reject transaction in aborted block.
+        """
+        self.__instances[tx_hash].status = TransactionStatus.REJECTED
+        self.__instances[tx_hash].block = None
+        self.__instances[tx_hash].transaction_failure_reason = TransactionFailureReason(
+            code=StarknetErrorCode.TRANSACTION_FAILED.name,
+            error_message="Block aborted.",
```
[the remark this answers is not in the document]
mikiw: I tried to find the message in the Pathfinder node but there is no such thing there (maybe it is in the sequencer?) so I made it up.

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    assert rpc_aborted_block_contract["result"]["status"] == "REJECTED"
+    rpc_aborted_block_invoke = rpc_call(
+        "starknet_getBlockWithTxs",
+        params={"block_id": {"block_hash": rpc_felt(invoke_block["block_hash"])}},
+    )
+    assert rpc_aborted_block_invoke["result"]["status"] == "REJECTED"
+
+    # Block and transaction should be accepted on L2
+    contract_deploy_info = deploy(CONTRACT_PATH, inputs=["0"])
```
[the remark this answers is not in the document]
mikiw: Fixed in 3965c31efc77337cb98c64357c4d5a28a85b9c4c

test/test_abort_blocks_after.py
mikiw: Fixed in a2be84291710ff1cce0287165dc7cab28a018351

starknet_devnet/starknet_wrapper.py
mikiw: Fixed in 4e0d79aa7ce421418e021dbd917652b22b7955b3

starknet_devnet/blueprints/base.py
mikiw: Fixed 82d227e5090031ced7f3aff2eabffcf5803b8173

mikiw:
Ok I think it's ready for re re review

FabijanC (changes requested):
I still need to review the test, but I don't trust GitHub's local storage that it will remain intact until tomorrow, so posting these comments now (and test review coming tomorrow).

starknet_devnet/starknet_wrapper.py
```diff
@@ -988,3 +987,48 @@ async def is_deployed(self, address: int) -> bool:
...
+            and self.config.fork_block >= starting_block.block_number
+            or starting_block.block_number == 0
+        ):
+            raise StarknetDevnetException(
+                code=StarknetErrorCode.OUT_OF_RANGE_BLOCK_ID,
+                message="Aborting genesis block or forked blocks is not supported.",
+            )
+
+        # Create new block with pending transactions if possible
+        if self.blocks.is_block_pending():
```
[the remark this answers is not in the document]
mikiw: After mining new block from pending block, transaction can be stored and can have status "REJECTED".

starknet_devnet/starknet_wrapper.py
```diff
@@ -988,3 +987,48 @@ async def is_deployed(self, address: int) -> bool:
...
+
+    async def abort_blocks(self, starting_block: StarknetBlock) -> str:
+        """
+        Abort blocks.
+        """
+        # Check if blocks can be aborted in fork mode
+        if (
+            self.config.fork_block
+            and self.config.fork_block >= starting_block.block_number
+            or starting_block.block_number == 0
```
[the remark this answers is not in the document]
mikiw: `starting_block.block_number == 0` is to prevent the abortion of genesis block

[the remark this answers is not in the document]
mikiw: we can do that, I'll change that

starknet_devnet/starknet_wrapper.py
```diff
@@ -988,3 +987,48 @@ async def is_deployed(self, address: int) -> bool:
...
+
+            aborted_blocks.append(hex(current_block.block_hash))
+            parent = await self.blocks.get_by_hash(hex(current_block.parent_block_hash))
+            current_block = parent
+
+        # Revert state
+        parent_block = await self.blocks.get_by_hash(
+            hex(starting_block.parent_block_hash)
+        )
+        self.starknet.state = self.blocks.get_state(parent_block.block_hash)
```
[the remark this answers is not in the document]
mikiw: but if we do it `get_state_update` by block_hash will start to fail

starknet_devnet/starknet_wrapper.py
```diff
@@ -988,3 +987,48 @@ async def is_deployed(self, address: int) -> bool:
...
+            )
+
+        # Create new block with pending transactions if possible
+        if self.blocks.is_block_pending():
+            await self.generate_latest_block()
+
+        aborted_blocks = []
+        current_block = await self.blocks.get_last_block()
+
+        # Abort blocks
```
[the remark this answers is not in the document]
mikiw: Fixed in e9180bef7dab32a04339c48febc6c63db692926c

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
FabijanC: I meant, "genesis block" instead of "origin block" (I edited the message). So my point is that genesis block isn't necessarily at 0.

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
FabijanC: You should add some kind of one-line comment to better motivate this.

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,297 @@
...
+
+@devnet_in_background(
+    *TESTNET_FORK_PARAMS, "--fork-block", str(TESTNET_DEPLOYMENT_BLOCK)
+)
+def test_forked_at_block_with_abort_blocks():
+    """Test GET on /fork_status when forking a given block"""
+    # Get latest valid block with block hash
+    fork_status = requests.get(f"{APP_URL}/fork_status")
+    assert fork_status.status_code == 200
+    last_block_by_number = get_block(
```
FabijanC, line 195: In what sense is this the last block by number as the variable name indicates?

test/test_abort_blocks_after.py
```diff
@@ -0,0 +1,297 @@
...
+    assert last_block_by_number["block_number"] == last_block["block_number"]
+    assert last_block_by_number["block_hash"] == last_block["block_hash"]
+    assert_tx_status(contract_deploy_info["tx_hash"], "ACCEPTED_ON_L2")
+
+
+@devnet_in_background(
+    *TESTNET_FORK_PARAMS, "--fork-block", str(TESTNET_DEPLOYMENT_BLOCK)
+)
+def test_forked_at_block_with_abort_blocks():
+    """Test GET on /fork_status when forking a given block"""
```
FabijanC, line 191: The description doesn't match the functionality of the test.

test/test_abort_blocks_after.py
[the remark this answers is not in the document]
FabijanC: I feel like there is a lot of similarity between this test and the previous couple of tests. What I meant is that there could be a test dedicated only to generating new blocks after aborting. So like: create a block, abort it, generate a new block, check that the new block is correct (without making other assertions since we already have them in other tests).

FabijanC (changes requested):
Review pt. 2: again raising concern about not testing the case of re-aborting the same block

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
mikiw: Fixed in 74b90cba85ac60e019b19ec787a2586f1052070b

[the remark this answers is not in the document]
FabijanC: In the commit that you linked, I see it only sets genesis_block_number if not forking (actually, if it's any falsy object; e.g. None and 0). Shouldn't it always be set?

Also, instead of adding more properties to starknet_wrapper, can this be added to the `DevnetBlocks` class?

mikiw:
re-aborting test case was added and some fixes were made

page/docs/guide/blocks.md
```diff
@@ -40,3 +40,27 @@ Response:
...
 {'block_hash': '0x115e1b390cafa7942b6ab141ab85040defe7dee9bef3bc31d8b5b3d01cc9c67'}
 ```
+
+### Abort blocks after
+
+This functionality allows to simulate block abortion that can occur on mainnet.
+
+You can abort blocks and reject transactions from the specified block to the currently latest block. Newly created blocks after the abortion will have accepted status and will continue with numbering where the last accepted block left off.
+
+Aborted blocks can only be queried by their hashes. The state of Devnet will be reverted to the last accepted block state. In a block on demand mode, the pending state will be reverted. Aborting the genesis block or blocks in forking origin is not supported.
```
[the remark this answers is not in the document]
mikiw: Fixed in 1a23fcfaab7bfb30bd8a8824e790d2720b05971c

test/test_abort_blocks_after.py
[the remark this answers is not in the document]
mikiw: True, fixed in 212db97a85408350386080013997291f10e18264

starknet_devnet/starknet_wrapper.py
```diff
@@ -988,3 +987,48 @@ async def is_deployed(self, address: int) -> bool:
...
+        if (
+            self.config.fork_block
+            and self.config.fork_block >= starting_block.block_number
+            or starting_block.block_number == 0
+        ):
+            raise StarknetDevnetException(
+                code=StarknetErrorCode.OUT_OF_RANGE_BLOCK_ID,
+                message="Aborting genesis block or forked blocks is not supported.",
+            )
+
```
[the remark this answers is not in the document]
mikiw: Fixed in 497a081ed99301d672bf49f506a5c3bf4a08fa71

starknet_devnet/starknet_wrapper.py
```diff
@@ -988,3 +987,48 @@ async def is_deployed(self, address: int) -> bool:
...
+        # Create new block with pending transactions if possible
+        if self.blocks.is_block_pending():
+            await self.generate_latest_block()
+
+        aborted_blocks = []
+        current_block = await self.blocks.get_last_block()
+
+        # Abort blocks
+        while current_block.block_number + 1 != starting_block.block_number:
```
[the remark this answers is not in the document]
mikiw: Fixed in c68e367ec46f98bc4b13019d37891c1b7f54ebaf

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
mikiw: Fixed in c68e367ec46f98bc4b13019d37891c1b7f54ebaf

starknet_devnet/blocks.py
```diff
@@ -168,14 +171,15 @@ async def get_state_update(
...
-                return self.__state_updates[block_number]
+            numeric_hash = self.__num2hash[block_number]
+            if numeric_hash in self.__state_updates:
+                return self.__state_updates[numeric_hash]
 
-            return await self.origin.get_state_update(block_number=block_number)
+            return await self.origin.get_state_update(block_hash=numeric_hash)
 
-        # now we know the block ID is "latest"
+        # now we get state updates by block_hash or from the origin
```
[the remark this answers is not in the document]
mikiw: Fixed

starknet_devnet/starknet_wrapper.py
```diff
@@ -617,14 +618,12 @@ async def __get_query_state(self, block_id: BlockId = DEFAULT_BLOCK_ID):
 
         assert isinstance(block_id, dict)
         if block_id.get("block_hash"):
-            # takes care of parsing
-            block = await self.blocks.get_by_hash(block_id["block_hash"])
-            return self.blocks.get_state(block.block_number)
+            return self.blocks.get_state(int(block_id.get("block_hash"), base=16))
```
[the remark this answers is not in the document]
mikiw: Fixed

test/test_abort_blocks_after.py
FabijanC: typo?

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
FabijanC: I meant that we can remove the old states, not the state updates (state updates should still work since we calculate them upon spawning a new block?)

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
FabijanC: From now on, let's refer to the first block in Devnet as `devnet_genesis_block`. Let's not allow aborting it (whether it is really a genesis (first block) or just the first block after forking).

starknet_devnet/blocks.py
```diff
@@ -177,7 +172,7 @@ async def get_state_update(
 
             return await self.origin.get_state_update(block_hash=numeric_hash)
 
-        # now we get state updates by block_hash or from the origin
+        # now an int
```
FabijanC, line 175: Isn't it latest here? Since the check before this checks if it's NOT latest

starknet_devnet/blocks.py
```diff
@@ -311,13 +306,20 @@ def get_state(self, block_hash: int) -> StarknetState:
...
     async def abort_block_by_hash(self, block_hash: str) -> str:
         """
         Abort block by given block hash.
         """
         numeric_hash = _parse_block_hash(block_hash)
 
-        if numeric_hash in self.__hash2block:
+        if numeric_hash in self.__hash2block and numeric_hash in list(
+            self.__num2hash.values()
+        ):
```
FabijanC, line 322:
```suggestion
        if numeric_hash in self.__hash2block and numeric_hash in self.__num2hash.values():
```

FabijanC: Wait actually, this will iterate through all `__num2hash` values for each block being aborted? I think a more efficient way is to have a dict of aborted blocks (e.g. `__hash2block_aborted`) and just check if the requested hash is not in there.

starknet_devnet/blocks.py
mikiw: LOL, got it, now I understand what you meant. Fixed in a40d6913d8d1842cb29473edfd3982b39ea76c99.

starknet_devnet/blocks.py
```diff
@@ -292,17 +296,36 @@ async def store_pending(
...
-        return self.__state_archive.get(block_number)
+        return self.__state_archive.get(block_hash)
+
+    async def abort_block_by_hash(self, block_hash: str) -> str:
+        """
+        Abort block by given block hash.
+        """
+        numeric_hash = _parse_block_hash(block_hash)
+
+        if numeric_hash in self.__hash2block:
```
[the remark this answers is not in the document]
FabijanC: You're right, the return value is used. But still, the function sometimes returns None. Perhaps it could raise an erroor instead? Also, now it seems to be related to https://github.com/0xSpaceShard/starknet-devnet/pull/421#discussion_r1177554595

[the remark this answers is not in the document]
FabijanC: Ahhh, we depend on it returning None to perform a check. In that case, the return type of the function should be `Optional[None]`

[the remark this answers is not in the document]
FabijanC: And we should explicitly return None. I really don't like the implicit return of None

mikiw:
I added a fix with exception instead of HTTP 200 e01c3ed20933c3509e19d91dca725423e0f69fe2

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
mikiw: Fixed in 8831b87f1a2ac2c867d30087f794974d6366114c

starknet_devnet/blocks.py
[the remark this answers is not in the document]
mikiw: Fixed in 98ba19b58fddda3bcc7c9ab5aebfa704643e7fd7

starknet_devnet/starknet_wrapper.py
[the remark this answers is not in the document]
mikiw: I'll try to change that and run all tests

starknet_devnet/blocks.py
mikiw: Check is one thing but the second thing is that we want to remove it later ```del self.__num2hash[block_dict["block_number"]]```.

starknet_devnet/starknet_wrapper.py
```diff
@@ -939,3 +942,71 @@ async def is_deployed(self, address: int) -> bool:
...
+
+        aborted_blocks = []
+        current_block = await self.blocks.get_last_block()
+
+        # Before the while loop to abort blocks, check if block numbers are set.
+        if not (current_block.block_number and starting_block.block_number):
+            raise StarknetDevnetException(
+                code=StarknetErrorCode.BLOCK_NOT_FOUND,
+                status_code=400,
+                message="Aborting blocks must have block numbers. This can happen during abortion of the same block twice.",
```
FabijanC, line 980: I think blocks needing a block number is an implementation detail.
```suggestion
                message="Block cannot be aborted. Make sure you are aborting an accepted block.",
```

mikiw: Fixed in 83173ec140027508606643845f52369c1f5da49e

FabijanC (approved):

PR #433, opened 2023-04-03
Fix continuation token showing after no events are left in get_events

### LINKED ISSUE #432: `continuation_token` appears in response body after no events are left in `get_events` RPC function
**Describe the bug (observed vs expected behavior)**

The `get_events` method in RPC doesn't match the specification. Parameter `continuation_token` appears in response body after no events searched for are left. 

After deploying a cairo contract with code:
```
%lang starknet
%builtins pedersen range_check

from starkware.cairo.common.cairo_builtins import HashBuiltin

@storage_var
func storage(key: felt) -> (value: felt) {
}

@event
func put_called(key: felt, prev_value: felt, value: felt) {
}

@external
func put{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(key: felt, value: felt) {
    let (prev_value) = storage.read(key);
    put_called.emit(key=key, prev_value=prev_value, value=value);
    storage.write(key, value);
    return ();
}
```
invoking `put` function 2 times and calling `get_events` in RPC with `"chunk_size": 4` the response looks like this:
```
{
    "id": 0,
    "jsonrpc": "2.0",
    "result": {
        "continuation_token": "1",
        "events": [
            {
                "block_hash": "0x06b0679b3060f2f1d02ec484f238f78a0a7c5aeb6bb0ced56edd40e8ddff05ff",
                "block_number": 3,
                "data": [
                    "0x01",
                    "0x00",
                    "0x01"
                ],
                "from_address": "0x021895a10d1f7319919b58967ec629d9fe6dc53269b0e24b8127ff4613759e55",
                "keys": [
                    "0x037e1449a2e98fcc818fa74e8abd5a3ebfefa48645be9b3777bd03e2c4ac54c0"
                ],
                "transaction_hash": "0x02304a74032d87b27f24537ae10105e29e014efb0fa9d0a9610132f490ac9373"
            },
            {
                "block_hash": "0x036f9108fb0b957df4d03eb18e7323c76c56219db854932a073d956ddba99347",
                "block_number": 4,
                "data": [
                    "0x02",
                    "0x00",
                    "0x02"
                ],
                "from_address": "0x021895a10d1f7319919b58967ec629d9fe6dc53269b0e24b8127ff4613759e55",
                "keys": [
                    "0x037e1449a2e98fcc818fa74e8abd5a3ebfefa48645be9b3777bd03e2c4ac54c0"
                ],
                "transaction_hash": "0x03037ea7104a49b4af996e0e3b21f886f1cd53c0742c8511d5f6ed2f553bec73"
            }
        ]
    }
}
```
Then, after adding `"continuation_token": "1"` to request body, the response looks like this:
```
{
    "id": 0,
    "jsonrpc": "2.0",
    "result": {
        "continuation_token": "1",
        "events": []
    }
}
```
I believe the problem lies [here](https://github.com/Shard-Labs/starknet-devnet/blob/v0.5.0a1/starknet_devnet/blueprints/rpc/misc.py#L155) and because [RpcEventsResult must return `continuation_token`, it isn't an optional parameter](https://github.com/Shard-Labs/starknet-devnet/blob/v0.5.0a1/starknet_devnet/blueprints/rpc/structures/responses.py#L79).

-----------------------------

**Not reproducible on alpha-goerli**

- [x] This issue is only present on Devnet and cannot be reproduced on alpha-goerli (check the box if true).

**To Reproduce**

1. Deploy any contract on devnet with function that emits an event.
2. Call the function that emits an event, e.g. 2 times.
3. Call `get_events` function in RPC and follow the `continuation_token` parameter. 
4. `continuation_token` appears in response body even though it shouldn't.

**Devnet version**

- I am using Devnet version: 0.5.0a1
- [ ] This happens with a dockerized Devnet (check the box if true).

tomek0123456789:
## Usage related changes

- `get_events` RPC endpoint works as declared in specification - `continuation_token` doesn't appear when there are no events left

Closes #432 

## Checklist:

- [x] Documented the changes
- [ ] Updated the tests
- [ ] All tests are passing - `./scripts/test.sh` (I couldn't set up the environment, however this function doesn't appear in tests)


starknet_devnet/blueprints/rpc/structures/responses.py
```diff
@@ -85,6 +85,14 @@ class RpcEventsResult(TypedDict):
     continuation_token: str
 
 
+class RpcEventsResultWithoutContinuationToken(TypedDict):
```
FabijanC, line 88: Can't we use class `RpcEventsResult` with continuation token being `Optional` (imported from `typing`)

FabijanC:
@tomek0123456789 Btw what about the tests, are they passing?

starknet_devnet/blueprints/rpc/structures/responses.py
tomek0123456789: I tried something like this: 
```
    if events:
        continuation_token = continuation_token + 1
        continuation_token = str(continuation_token)
    else:
        continuation_token = None

    return RpcEventsResult(events=events, continuation_token=continuation_token)
```
and
```
class RpcEventsResult(TypedDict):
    """
    TypedDict for rpc get events result
    """

    events: List[EmittedEvent]
    continuation_token: Optional[str]
```
but then when I called an event 4 times, then called `get_events` with `"chunk_size": 1` and followed continuation token, when called with `"continuation_token": "4"` devnet returned INTERNAL_ERROR code (-32603)
image

cptartur (commented):
Looks fine to me. I'd verify that we have a good test coverage for these changes, but otherwise it's good.

FabijanC:
@tomek0123456789 Can you confirm that tests are passing? Also, can you improve the testing of this feature as @cptartur suggested? I was hoping we can release today with this PR and #435 getting merged.

test/rpc/test_rpc_misc.py
```diff
@@ -200,6 +201,59 @@ def test_get_events_wrong_blockid_type():
...
+    deploy_info = deploy(EVENTS_CONTRACT_PATH)
+    for i in range(3):
+        invoke(
+            calls=[(deploy_info["address"], "increase_balance", [i])],
+            account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
+            private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+        )
+    resp = rpc_call(
+        "starknet_getEvents",
+        params=create_get_events_rpc(create_block_from_0(3))["params"],
```
FabijanC, line 223: I see we are using the `create_get_events_rpc` function only to extract the `params` from it (also present in some old tests). This looks like a problem, and should be dealt with. If you see a simple solution that won't require several hours of refactoring, you can do it in this PR, otherwise I will create a new GitHub issue. But I need your feedback, please.

test/rpc/test_rpc_misc.py
```diff
@@ -200,6 +201,59 @@ def test_get_events_wrong_blockid_type():
     assert resp["error"]["code"] == PredefinedRpcErrorCode.INVALID_PARAMS.value
 
 
+@pytest.mark.usefixtures("run_devnet_in_background")
+@pytest.mark.parametrize(
+    "run_devnet_in_background",
+    [PREDEPLOY_ACCOUNT_CLI_ARGS],
+    indirect=True,
+)
```
FabijanC, line 209: Can be simplified as

```suggestion
@devnet_in_background(*PREDEPLOY_ACCOUNT_CLI_ARGS)
```

test/rpc/test_rpc_misc.py
```diff
@@ -200,6 +201,59 @@ def test_get_events_wrong_blockid_type():
...
+def test_get_events_continuation_token():
+    """
+    Test RPC get_events returning continuation token.
+    """
+    deploy_info = deploy(EVENTS_CONTRACT_PATH)
+    for i in range(3):
```
FabijanC, line 215: Can you extract this 3 to a variable of this function (something like `total_invokes`)? Then use it where applicable.

test/rpc/test_data/get_events.py
```diff
@@ -91,27 +93,6 @@
     "chunk_size": 10,
 }
 
-BLOCK_FROM_0_TO_LATEST_CHUNK_SIZE_1 = {
```
FabijanC, line 94: Thank you for removing at least a part of these inconvenient constants xD

FabijanC (changes requested):
Thanks for the thorough test!

test/rpc/test_rpc_misc.py
cptartur: If it does require extensive refactoring we could at least define a separate method that extracts this params and use it here. So it will be easier to replace eventually.

test/rpc/test_rpc_misc.py
```diff
@@ -200,6 +201,59 @@ def test_get_events_wrong_blockid_type():
...
+    )
+    assert len(resp["result"]["events"]) == 1
+    assert resp["result"]["continuation_token"] == "2"
+
+    resp = rpc_call(
+        "starknet_getEvents",
+        params=create_get_events_rpc(create_block_from_0(1, "2"))["params"],
+    )
+    assert len(resp["result"]["events"]) == 1
+    assert "continuation_token" not in resp["result"]
```
cptartur, line 247: Wonder if using `@pytest.mark.parametrize` instead of multiple lines of basically duplicated code would make it more readable.

test/rpc/test_data/get_events.py
tomek0123456789: I think I can change the method creating the `filter` parameter to have all the parameters needed to create those constants, what do you think?

tomek0123456789:
After bumping cairo-lang version, one test started to fail
```
    @pytest.mark.usefixtures("devnet_with_account")
    def test_get_state_update():
        """Test if declared classes successfully registered"""
    
        declare_info = declare(
            contract_path=CONTRACT_PATH,
            account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
            private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
            max_fee=int(4e16),
        )
        contract_class_hash = declare_info["class_hash"]
>       assert_hex_equal(contract_class_hash, EXPECTED_CLASS_HASH)

test/rpc/test_rpc_misc.py:59: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
358fb209d5b7d0b1558a6163858071a55148b', expected = '0x2216379233d6f7c8890a730e164e069dafdd63edf0f0218dde8181b50fabd45'

```

FabijanC:
> After bumping cairo-lang version, one test started to fail
> 
> ```
>     @pytest.mark.usefixtures("devnet_with_account")
>     def test_get_state_update():
>         """Test if declared classes successfully registered"""
>     
>         declare_info = declare(
>             contract_path=CONTRACT_PATH,
>             account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
>             private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
>             max_fee=int(4e16),
>         )
>         contract_class_hash = declare_info["class_hash"]
> >       assert_hex_equal(contract_class_hash, EXPECTED_CLASS_HASH)
> 
> test/rpc/test_rpc_misc.py:59: 
> _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
> 358fb209d5b7d0b1558a6163858071a55148b', expected = '0x2216379233d6f7c8890a730e164e069dafdd63edf0f0218dde8181b50fabd45'
> ```

Can you rerun the contract-compilation script `scripts/compile_contracts.sh`?

test/rpc/test_rpc_misc.py
```diff
@@ -217,21 +260,12 @@ def test_get_events(input_data, expected_data):
             account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
             private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
         )
-    resp = rpc_call("starknet_getEvents", params=input_data["params"])
+    resp = rpc_call("starknet_getEvents", params=input_data)
+    print(resp)
```
FabijanC, line 264: probably an unwanted print

test/rpc/test_data/get_events.py
```diff
@@ -8,120 +8,65 @@
...
-    "from_block": {"block_number": 2},
-    "to_block": {"block_number": 3},
-    "address": rpc_felt(EXPECTED_FEE_TOKEN_ADDRESS),
-    "keys": [
-        rpc_felt(FEE_CHARGED_EVENT_KEY),
-    ],
-    "chunk_size": 10,
-}
+def parse_block_delimiter_parameter(
+        block_delimiter: Union[str, int]
```
FabijanC, line 55: There is a block identifier type in starkware.starknet.services.api.feeder_gateway.response_objects, can that be used here?

test/rpc/test_data/get_events.py
```diff
@@ -8,120 +8,65 @@
...
+    *,
+    from_block: Union[int, str] = 0,
+    to_block: Union[int, str] = "latest",
+    address: Optional[str] = rpc_felt(EXPECTED_FEE_TOKEN_ADDRESS),
+    keys: Optional[List[str]] = None,
+    chunk_size: int = 10,
+    continuation_token: Optional[str] = None
+) -> dict:
+    """
+    Method that creates ``filter`` parameter for `get_events` RPC call.
```
FabijanC, line 26: This is more docstring than I would expected, thanks for that! Probably unimportant, but for it's just that some of the optional parameters indicate that they are optional while others don't.

FabijanC (changes requested):
Looks like big improvement, great! Just left some minor comments.

test/rpc/test_data/get_events.py
cptartur: I think if we have a default value that is not `None` we shouldn't be using `Optional`. It looks okay to me.

test/rpc/test_rpc_misc.py
```diff
@@ -193,13 +183,66 @@ def test_get_events_wrong_blockid_type():
...
+    Test RPC get_events returning continuation token.
+    """
+    deploy_info = deploy(EVENTS_CONTRACT_PATH)
+    total_invokes = 3
+    for i in range(total_invokes):
+        invoke(
+            calls=[(deploy_info["address"], "increase_balance", [i])],
+            account_address=PREDEPLOYED_ACCOUNT_ADDRESS,
+            private_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+        )
```
cptartur, line 207: How about making it a separate fixture defined in this file?

test/rpc/test_rpc_misc.py
```diff
@@ -193,13 +183,66 @@ def test_get_events_wrong_blockid_type():
...
+    resp = rpc_call(
+        "starknet_getEvents",
+        params=create_get_events_filter(chunk_size=total_invokes),
+    )
+    assert len(resp["result"]["events"]) == total_invokes
+    assert "continuation_token" not in resp["result"]
```
cptartur, line 213: This could probably be refactored into a separate method with signature like
```py
def assert_no_continuation_token_present(chunk_size: int):
  ...
```

cptartur: Or even more generic
```py
def assert_get_events_response(chunk_size: int, expected_token: Optional[str], continuation_token: Optional[str] = None):
  """
  If expected_token is None, check that it is not present, if it's not None, check if the returned value matches 
  provided.
  """
```

Then you could probably replace all of the code below with calls to this method.

test/rpc/test_data/get_events.py
FabijanC: `Optional[...]` is a shorthand notation for `Union[..., None]` so it might make sense to have it like this if it can be `None` (if we are talking about `address`).
[Source](https://stackoverflow.com/a/51710151)

tomek0123456789: Used `Optional` for address because of one fixture in one test (particularly for `test_get_events_missing_parameter`), however now I think it isn't necessary - the dictionary can be just changed after calling the function (like in test below that one - `test_get_events_wrong_blockid_type`). Tell me what you prefer.

test/rpc/test_data/get_events.py
tomek0123456789: I think it works, for the sake of clarity I'd make that something like `Union[BlockIdentifier, BlockHash]`, where `BlockHash = str`.

FabijanC: Right, now I see we can't just use `BlockIdentifier` since we want to support hashes. Regarding `BlockHash`, that type is defined somewhere (rpc/structures/types.py)

test/rpc/test_data/get_events.py
FabijanC: Modifying the dict is better IMO.

test/rpc/test_rpc_misc.py
tomek0123456789: Not sure how to do this. Is it possible to run devnet on a fixture, like with `@devnet_in_background(*PREDEPLOY_ACCOUNT_CLI_ARGS)`? I marked the fixture with it and marked `test_get_events_continuation_token` with `@pytest.mark.usefixtures("invoke_increase_balance")` - the test then gives an error 
```
E           requests.exceptions.ConnectionError: HTTPConnectionPool(host='127.0.0.1', port=49990): 
Max retries exceeded with url: /gateway/add_transaction (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x119fe99d0>:
 Failed to establish a new connection: [Errno 61] Connection refused'))
```
@FabijanC

FabijanC: Maybe Artur was refering to something like the existing "deploy_info" and "invoke_info" in `test/rpc/conftest.py` where you could add transactions that you need.

But IMO it's ok to keep the invokes explicitly here in the body of the test.

test/rpc/test_rpc_misc.py
tomek0123456789: Done like suggested below

FabijanC (approved):

PR #442, opened 2023-04-12
Internal error on get_full_contract

mikiw:
## Usage related changes

- Bugfix of UDC deployment nad get_full_contract()

- Fix missing set of contract_classes for UDC invoke

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/starknet_devnet/starknet_wrapper.py
+++ b/starknet_devnet/starknet_wrapper.py
@@ -1,4 +1,3 @@
-# pylint: disable=too-many-lines
 """
 This module introduces `StarknetWrapper`, a wrapper class of
 starkware.starknet.testing.starknet.Starknet.
@@ -595,7 +594,6 @@ class StarknetWrapper:
     async def invoke(self, external_tx: InvokeFunction):
         """Perform invoke according to specifications in `transaction`."""
         state = self.get_state()
-
         async with self.__get_transaction_handler(
             external_tx=external_tx
         ) as tx_handler:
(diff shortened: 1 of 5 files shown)


starknet_devnet/starknet_wrapper.py
```diff
@@ -734,7 +734,11 @@ async def get_class_by_address(
         state = await self.__get_query_state(block_id)
         cached_state = state.state
         class_hash = await self.get_class_hash_at(contract_address)
-        return cached_state.contract_classes[class_hash]
+
+        if class_hash in cached_state.contract_classes:
+            return cached_state.contract_classes[class_hash]
+
+        return self._contract_classes[class_hash]
```
FabijanC, line 741: Can this fail? I think `self.get_class_hash_at` (done two lines above), will default to polling origin and fail correctly if the contract is not deployed on devnet or forking origin. But what if it's deployed on origin (meaning `self.get_class_hash_at` succeeds), but the class does not exist locally (meaning no key in `contract_classes`)? If this is indeed an issue, it can be addressed in a separate issue/PR.

FabijanC (changes requested):

starknet_devnet/starknet_wrapper.py
mikiw: I was thinking about that but that would mean that we have an error in another place in the code right?

mikiw: So if the contract is forked it will be in `cached_state.contract_classes` or `self._contract_classes`?

mikiw: I think that Clement from kakarot team had issue with `get_full_contract` and forking

FabijanC (approved):

PR #462, opened 2023-05-08
Wrap badrequest error when forking

### LINKED ISSUE #423: Wrap BadRequest error logs when forking
When defaulting to the forked network, if the requested object is missing (e.g. an undeclared class), the feeder gateway client logs the error (without actually raising) - this is not a bug, but may be misleading.

This should be handled in origin.py; we could use the same mechanism used in ForkedState (and apply it in other places in ForkedState where missing)

tonypony220:
## Usage related changes
- Closes #423

Not wrapped functions in Forked state seems to have no errors and not have to be suppressed.

- none

- redirect stderr replaced with muting only logger (not redirects in origin for some reason). 

## Checklist:

- [ ] Documented the changes
- [x] Updated the tests

### DIFF
--- a/starknet_devnet/util.py
+++ b/starknet_devnet/util.py
@@ -1,10 +1,10 @@
 """
 Utility functions used across the project.
 """
+import importlib.util
 import logging
 import os
 import sys
-import contextlib
 from dataclasses import dataclass
 from typing import Dict, List, Set, Tuple
 
@@ -247,6 +247,8 @@ def warn(msg: str, file=sys.stderr):
 
 
 class SuppressLogger:
+    """ctx mng to suppress logger"""
+
     def __init__(self, logger):
         self.logger = logger
 
@@ -256,7 +258,7 @@ class SuppressLogger:
     def __exit__(self, exc_type, exc_val, exc_tb):
         self.logger.disabled = False
 
-import importlib.util
+
 client_spec = importlib.util.find_spec("services.external_api.client")
 assert client_spec
 
--- a/starknet_devnet/util.py
+++ b/starknet_devnet/util.py
@@ -1,7 +1,6 @@
 """
 Utility functions used across the project.
 """
-import importlib.util
 import logging
 import os
 import sys
@@ -246,11 +245,13 @@ def warn(msg: str, file=sys.stderr):
     print(f"\033[93m{msg}\033[0m", file=file)
 
 
-class SuppressLogger:
-    """ctx mng to suppress logger"""
+class LogSuppressor:
+    """Context manager to suppress logger"""
 
-    def __init__(self, logger):
-        self.logger = logger
+    def __init__(self, logger_name):
+        # check logger exists
+        assert logger_name in logging.Logger.manager.loggerDict
+        self.logger = logging.getLogger(logger_name)
 
     def __enter__(self):
         self.logger.disabled = True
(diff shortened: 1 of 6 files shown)


starknet_devnet/util.py
```diff
@@ -243,3 +244,26 @@ def get_fee_estimation_info(tx_fee: int, gas_price: int):
...
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.logger.disabled = False
+
+
+client_spec = importlib.util.find_spec("services.external_api.client")
+assert client_spec
+
+# FeederGatewayClient is implemented in such a way that it logs and raises;
+# this suppresses the logging
+suppress_feeder_gateaway_client_logger = SuppressLogger(
```
FabijanC, line 267: What's the difference between this and `contextlib.redirect_stderr(None)` that has been used so far?

test/test_endpoints.py
FabijanC, line 1: Why was it necessary to update the tests?

starknet_devnet/util.py
tonypony220: `contextlib.redirect_stderr(None)` somehow not works when used in ForkedOrigin methods.

test/test_endpoints.py
tonypony220: Not necessarily in general, just convenient to check. i will then discard the changes checking the wrapper

starknet_devnet/util.py
```diff
@@ -243,3 +244,26 @@ def get_fee_estimation_info(tx_fee: int, gas_price: int):
...
+        self.logger = logger
+
+    def __enter__(self):
+        self.logger.disabled = True
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        self.logger.disabled = False
+
+
+client_spec = importlib.util.find_spec("services.external_api.client")
```
FabijanC, line 262: Why is this imported like this?
What about
```python
import services.external_api.client as client_spec
```

And what even is this `services` module? I see it's not native to Python, but it's also not a dependency listed in `pyproject.toml`.

starknet_devnet/util.py
```diff
@@ -243,3 +244,26 @@ def get_fee_estimation_info(tx_fee: int, gas_price: int):
 def warn(msg: str, file=sys.stderr):
     """Log a warning"""
     print(f"\033[93m{msg}\033[0m", file=file)
+
+
+class SuppressLogger:
+    """ctx mng to suppress logger"""
```
FabijanC, line 250: Can you use full words here? Is that supposed to be `Context manager`?

starknet_devnet/util.py
```diff
@@ -243,3 +244,26 @@ def get_fee_estimation_info(tx_fee: int, gas_price: int):
 def warn(msg: str, file=sys.stderr):
     """Log a warning"""
     print(f"\033[93m{msg}\033[0m", file=file)
+
+
+class SuppressLogger:
```
FabijanC, line 249: Class names shouldn't be verbs but nouns (with possible descriptors). E.g. `SuppressedLogger` or `LogSuppressor`.

FabijanC (changes requested):

starknet_devnet/util.py
tonypony220: Yes, my bad. Rewrote to check the existence of the target logger, not the module. 
The `services` module is associated with Everest, from which FeederGatewayClient is inherited.

starknet_devnet/util.py
tonypony220: fixed

FabijanC (approved):

PR #472, opened 2023-05-16
 0.11.1 adaptations

mikiw:
## Usage related changes

- Adaptations to cairo-lang v0.11.1.1
- DEFAULT_GAS_PRICE was set to 10**8 - the same as in starkware codebase

## Checklist:

- [x] Documented the changes
- [x] Updated the tests


starknet_devnet/starknet_wrapper.py
```diff
@@ -329,7 +329,7 @@ async def update_pending_state(  # pylint: disable=too-many-arguments
         )
 
         return BlockStateUpdate(
-            block_hash=None,
+            block_hash=0,
```
mikiw, line 332: I'm not sure about this line.

mikiw:
```
@marshmallow_dataclass.dataclass(frozen=True)
class BlockStateUpdate(ValidatedResponseObject):
    """
    Represents a response block state update.
    """

    block_hash: Optional[int] = field(metadata=fields.optional_block_hash_metadata)
    new_root: Optional[bytes] = field(metadata=fields.optional_state_root_metadata)
    old_root: bytes = field(metadata=fields.state_root_metadata)
    state_diff: StateDiff

    def __post_init__(self):
        super().__post_init__()
        assert (self.block_hash is None) == (
            self.new_root is None
        ), "new_root must appear in state update for any block other than pending block."
```

mikiw: so we have this assert `assert (self.block_hash is None) == (self.new_root is None)`

starknet_devnet/postman_wrapper.py
```diff
@@ -199,7 +198,6 @@ def __init__(
 
         # Create a filter to collect LogMessageToL2 events.
         w3_contract = self.mock_starknet_messaging_contract.w3_contract
-        web3_contract_create_filter_fix(w3_contract.events.LogMessageToL2)
```
FabijanC, line 202: What happened here? Wasn't this filter fix important?

starknet_devnet/general_config.py
```diff
@@ -34,7 +33,6 @@ def build_devnet_general_config(chain_id: StarknetChainId):
                 "chain_id": chain_id.value,
                 "fee_token_address": hex(FeeToken.ADDRESS),
             },
-            "tx_version": SUPPORTED_TX_VERSION,
```
FabijanC, line 37: is tx_version now specified elsewhere?

starknet_devnet/starknet_wrapper.py
FabijanC: So block_hash and new_root need to be of the same `None`ness. Is it legal if all three values are `None` (block hash, old root, new root)? If not, maybe then either add a comment explaining that this is a dummy value, or create a new constant (similar to DUMMY_ROOT)

starknet_devnet/constants.py
```diff
@@ -13,7 +13,7 @@
 
 DEFAULT_ACCOUNTS = 10
 DEFAULT_INITIAL_BALANCE = 10**21
-DEFAULT_GAS_PRICE = 10**11
```
FabijanC, line 16: I think this ought to be documented in the PR description usage changes. And run.md might has to be updated, too.

FabijanC (changes requested):

starknet_devnet/postman_wrapper.py
mikiw: This function was removed, maybe they changed something and it's not needed anymore.

I remember when I commented that during another L1L2 development tests were failing so it was needed and now without it they are passing so it's ok.

starknet_devnet/constants.py
mikiw: added

mikiw: bbb8be7c6e73f8eb55148e5f596c093f420109ce

starknet_devnet/general_config.py
mikiw: I'll check but it was removed from `StarknetGeneralConfig` - that's for sure

FabijanC: it's ok, I guess it's not really important since specifying the supported version wasn't something that users could modify.

starknet_devnet/starknet_wrapper.py
mikiw: Validation is not passing if `new_root` or `block_hash` is not set (strange because it's optional). And both can't be `None` so I just will add `DUMMY_PENDING_BLOCK_HASH`.

mikiw:
after merge test are failing, I'm on it

FabijanC (approved):

PR #489, opened 2023-05-29
bugfix for __postman_wrapper on init

mikiw:
## Usage related changes

- bugfix for `flush` endpoint call without previous call of `load_l1_messaging_contract`

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/starknet_devnet/postman_wrapper.py
+++ b/starknet_devnet/postman_wrapper.py
@@ -106,7 +106,7 @@ and that the Messaging Contract is deployed at the provided address ({contract_a
         """Handles all pending L1 <> L2 messages and sends them to the other layer."""
 
         if self.__postman_wrapper is None:
-            return ([],[])
+            return ([], [])
 
         postman = self.__postman_wrapper.postman
 
--- a/starknet_devnet/postman_wrapper.py
+++ b/starknet_devnet/postman_wrapper.py
@@ -106,7 +106,7 @@ and that the Messaging Contract is deployed at the provided address ({contract_a
         """Handles all pending L1 <> L2 messages and sends them to the other layer."""
 
         if self.__postman_wrapper is None:
-            return ([], [])
+            return ({}, [])
 
         postman = self.__postman_wrapper.postman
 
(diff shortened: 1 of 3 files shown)


starknet_devnet/postman_wrapper.py
```diff
@@ -106,7 +106,7 @@ async def flush(self, state) -> Tuple[dict, list]:
         """Handles all pending L1 <> L2 messages and sends them to the other layer."""
 
         if self.__postman_wrapper is None:
-            return {}
+            return ([], [])
```
FabijanC, line 109: Can be `return [], []` but you can leave it as it is if you think it's clearer that way.

starknet_devnet/starknet_wrapper.py
```diff
@@ -755,24 +755,25 @@ async def mock_message_to_l2(self, transaction: InternalL1Handler) -> dict:
         return transaction.hash_value
 
     async def postman_flush(self) -> dict:
-        """Handles all pending L1 <> L2 messages and sends them to the other layer."""
+        """Handles all pending L1 <> L2 messages if such exists and sends them to the other layer."""
```
FabijanC, line 758: I don't see these three new words as really necessary. In fact, the current wording is grammatically incorrect.

FabijanC (changes requested):
As discussed: can be a simple test which just attempts flushing before anything has been done previously postman-wise. And we should just assert nothing has failed. Maybe we don't even have to load anything in this new test.

starknet_devnet/postman_wrapper.py
mikiw: I'll leave it as it is

starknet_devnet/starknet_wrapper.py
mikiw: rollbacked in 2eff0d05672cf00b5a63de88b6fc809ce5df205e

mikiw:
I added new test

FabijanC (approved):

PR #500, opened 2023-06-14
Fix `rpc_declare_transaction` logic to also return DeclareV2

tomek0123456789:
## Usage related changes

- `getBlockWithTxs`, `getBlockWithTxHashes` and `getTransactionByHash` now can return `DeclareV2` transaction

- changed `rpc_declare_transaction` function to match RPC spec

## Checklist:

- [x] Documented the changes
- [x] Updated the tests

### DIFF
--- a/starknet_devnet/blueprints/rpc/structures/payloads.py
+++ b/starknet_devnet/blueprints/rpc/structures/payloads.py
@@ -7,7 +7,7 @@ RPC payload structures
 from __future__ import annotations
 
 from enum import Enum, auto
-from typing import Callable, Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Union, cast
 
 from marshmallow.exceptions import MarshmallowError
 from starkware.starknet.definitions.constants import QUERY_VERSION_BASE
@@ -226,6 +226,11 @@ class RpcDeclareTransaction(RpcTransactionCommon):
     sender_address: Address
 
 
+class RpcDeclareV2Transaction(RpcDeclareTransaction):
+
+    compiled_class_hash: Felt
+
+
 class RpcDeployTransaction(TypedDict):
     """TypedDict for rpc deploy transaction"""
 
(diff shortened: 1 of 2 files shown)


starknet_devnet/blueprints/rpc/structures/payloads.py
```diff
@@ -330,6 +336,13 @@ def rpc_declare_transaction(transaction: DeclareSpecificInfo) -> RpcDeclareTrans
         "nonce": rpc_felt(transaction.nonce),
         "type": rpc_txn_type(transaction.tx_type.name),
     }
+    if transaction.version == 2:
```
FabijanC, line 339: Let's not use the hardcoded value here. We have constants for this in `starknet_devnet/constants.py`.

test/rpc/test_rpc_transactions.py
```diff
@@ -181,6 +182,46 @@ def test_get_transaction_by_hash_declare():
...
+        contract_class=contract_class,
+        compiled_class_hash=compiled_class_hash,
+        sender_address=PREDEPLOYED_ACCOUNT_ADDRESS,
+        sender_key=PREDEPLOYED_ACCOUNT_PRIVATE_KEY,
+    )
+    assert_declare_v2_accepted(declaration_resp)
+
+    declare_info = declaration_resp.json()
+    block = get_block_with_transaction(declare_info["tx_hash"])
+    block_tx = block["transactions"][0]
```
FabijanC, line 202: Before extracting the value from index 0, can you assert that len is 1?

starknet_devnet/blueprints/rpc/structures/payloads.py
```diff
@@ -226,6 +226,12 @@ class RpcDeclareTransaction(RpcTransactionCommon):
     sender_address: Address
 
 
+class RpcDeclareV2Transaction(RpcDeclareTransaction):
```
FabijanC, line 229: Should these types be framed in the same way we already have it with broadcasted types (notice the V1, V2 and Union):
https://github.com/0xSpaceShard/starknet-devnet/blob/4a8d028c5babea9616041781d21e337b55d7936d/starknet_devnet/blueprints/rpc/structures/payloads.py#L155-L170

FabijanC (changes requested):

